PR #72, opened 2022-07-05
WIP adding oscn_scraper link parser to oscn

AspiringDev51:
WIP: Issue with court_cases.rake. data variable doesn't hold information expected. doesn't have a link with parameters for county court case number or oscn number. reading the file It seems its gives an error page.


holdenmitchell (approved):
👍

PR #74, opened 2022-07-11
improved spec for new_cases_spec

AspiringDev51:
Improving new_case_spec

### DIFF
--- a/app/services/importers/new_court_case.rb
+++ b/app/services/importers/new_court_case.rb
@@ -24,7 +24,7 @@ module Importers
     private
 
     def oscn_id
-      params['casemasterid'].first.to_i
+      params['casemasterID'].first.to_i
     end
 
     def county
--- a/spec/services/importers/new_cases_spec.rb
+++ b/spec/services/importers/new_cases_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe Importers::NewCourtCase do
       expect(court_case.county.name).to eq 'Oklahoma'
     end
 
-    it ' does not duplicate' do
+    it 'does not duplicate' do
       county = create(:county, name: 'Oklahoma')
       create(:court_case, :felony, county_id: county.id, oscn_id: 3_946_802)
       file_path = 'spec/fixtures/importers/link.html'
--- a/spec/services/importers/new_cases_spec.rb
+++ b/spec/services/importers/new_cases_spec.rb
@@ -2,8 +2,23 @@ require 'rails_helper'
 
 RSpec.describe Importers::NewCourtCase do
   describe '#perform' do
-    it 'add specs' do
-      skip
+    it 'perform correctly finds or creates and returns a new case' do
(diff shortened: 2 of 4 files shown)


spec/services/importers/new_cases_spec.rb
```diff
@@ -2,8 +2,21 @@
...
+      file_path = 'spec/fixtures/importers/link.html'
+      html = File.read(file_path)
+      link_html = Nokogiri::HTML.parse(html)
+      link_html = link_html.css('a').first
+
+      data = described_class.new(link_html)
+      data.perform
+      expect(data.case_number).to eq 'CF-2021-489'
+      expect(data.case_types.keys.first).to eq 'CF'
+      expect(data.counties.keys.first).to eq 'Oklahoma'
```
holdenmitchell, line 19: I think we want to directly query the database to make sure that the data has been saved. It should be saved, but if the `Importers::NewCourtCase.perform` method were to return the object without saving to the DB, this test would falsely pass

```ruby
court_case = CourtCase.find_by(county_id: county.id, oscn_id: oscn_id)

expect(court_case.case_number) to eq 'CF-2021-489'
# ...
```

spec/services/importers/new_cases_spec.rb
```diff
@@ -2,8 +2,21 @@
...
     end
```
holdenmitchell, line 20: I'm thinking there could be two different specs here:

1) Test that it creates the case when it does not exist in the database (What you currently have)
2) Test that it does not duplicate a case if it already exists in the database (Different spec)

holdenmitchell (changes requested):
Left some small change requests! 🥈

spec/services/importers/new_cases_spec.rb
```diff
@@ -2,8 +2,37 @@
...
+      data.perform
+      oscn_id = data.params['casemasterID']
+      court_case = CourtCase.find_by(county_id: county.id, oscn_id: oscn_id)
+
+      expect(court_case.case_number).to eq 'CF-2021-489'
+      expect(court_case.case_number[0..1]).to eq 'CF'
+      expect(court_case.county.name).to eq 'Oklahoma'
+    end
+
+    it ' does not duplicate' do
```
sgelbart, line 24:
```suggestion
    it 'does not duplicate' do
```

sgelbart (approved):

app/services/importers/new_court_case.rb
```diff
@@ -24,7 +24,7 @@ def perform
     private
 
     def oscn_id
-      params['casemasterid'].first.to_i
+      params['casemasterID'].first.to_i
```
sgelbart, line 27: did you mean to update this?

AspiringDev51: Yea, the param name was wrong, needed to be casemasterID

holdenmitchell (approved):

PR #76, opened 2022-07-11
improving medium_priority_spec

AspiringDev51:
improving medium_priority_spec

Test via` rspec .`

### DIFF
--- a/spec/services/scrapers/medium_priority_spec.rb
+++ b/spec/services/scrapers/medium_priority_spec.rb
@@ -10,10 +10,9 @@ RSpec.describe Scrapers::MediumPriority do
       false_case = create(:court_case, :with_html, closed_on: '6/11/2022')
       false_case.case_html.update(scraped_at: '7/11/2022')
 
-      # data = described_class.perform
       expect do
-        described_class.perform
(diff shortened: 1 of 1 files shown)


spec/services/scrapers/medium_priority_spec.rb
```diff
@@ -1,9 +1,18 @@
...
 
 RSpec.describe Scrapers::MediumPriority do
   describe '#perform' do
-    it 'add specs' do
-      skip
+    it 'adds jobs to the CourtCaseWorker' do
+      test_case = create(:court_case, :with_html, closed_on: '6/15/2022')
+      test_case.case_html.update(scraped_at: '6/15/2022')
+      false_case = create(:court_case, :with_html, closed_on: '6/11/2022')
+      false_case.case_html.update(scraped_at: '7/11/2022')
```
holdenmitchell, line 11: Similar to the [other PR comment](https://github.com/AyOK-Code/oscn/pull/77/files#r920453990), you'll need to take into account the date that the test is being run on. As time passes, this spec will start to fail. See other PR for solutions

holdenmitchell (changes requested):
I missed this on the first pass! Quick change and then it'll be ready

PR #77, opened 2022-07-11
high_priority_spec improved

AspiringDev51:
high_priority_spec improved

Test with rspec .

### DIFF
--- a/spec/services/scrapers/high_priority_spec.rb
+++ b/spec/services/scrapers/high_priority_spec.rb
@@ -6,9 +6,9 @@ RSpec.describe Scrapers::HighPriority do
   describe '#perform' do
     it 'add specs' do
       county = create(:county, name: 'Tulsa')
-      test_case = create(:court_case, closed_on: '6/15/2022', county: county)
+      test_case = create(:court_case, closed_on: 7.days.ago, county: county)
 
-      create(:court_case, closed_on: '7/11/2022', county: county)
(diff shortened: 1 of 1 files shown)


spec/services/scrapers/high_priority_spec.rb
```diff
@@ -1,9 +1,20 @@
...
+Sidekiq::Testing.fake!
 
 RSpec.describe Scrapers::HighPriority do
   describe '#perform' do
     it 'add specs' do
-      skip
+      county = create(:county, name: 'Tulsa')
+      test_case = create(:court_case, closed_on: '6/15/2022', county: county)
+
+      create(:court_case, closed_on: '7/11/2022', county: county)
```
holdenmitchell, line 11: This spec is going to start failing in ~5 days because the dates are statically coded in. Once the date `7/11/2022`, 

You have a couple of options here:

1) Create the dates relative to the date that the test is being run on:

```ruby
create(:court_case, closed_on: 2.days.ago, county: county)
```

2) Try using the `travel_to` helper [described here](https://til.hashrocket.com/posts/846d1b9d55-time-travelling-in-rspecrails). This will make it so the spec travels to a date and then runs as if it is that date.

holdenmitchell (changes requested):
See comment about the statically coded dates

holdenmitchell (approved):
👍

PR #78, opened 2022-07-13
WIP Case spec improvement

AspiringDev51:
WIP Case spec improvement


holdenmitchell:
@AspiringDev51 What specifically are you looking for help with on this PR?

AspiringDev51:
> @AspiringDev51 What specifically are you looking for help with on this PR?


@holdenmitchell  For this PR Im trying to the case.rb file. it seems to run the low,med,and high priority files, but I'm having trouble figuring out exactly what I should be testing for, as case.rb's purpose is a bit hard for me to understand. does it simply run the files and return the cases? or something else?

PR #88, opened 2022-07-25
issue 82

AspiringDev51:
**To Test**
run `rails console`
 Rails.application.load_tasks 
Rake::Task['doc:doc_facilities'].invoke

~~**Problem With Duplicates**~~
~~ran the task and noticed that there are many repeat statuses that didn't get their associated  doc_facility_id. ~~

~~using LAWTON CASELOAD (BECKHAM-CUSTER) as an example we see that in total we have 1135 statuses with that facility. 3 of those have  doc_facility_ids~~

~~on a larger scale,  there are 236 statuses that have an id.~~

~~removing the duplicate status gets us 204~~

~~those 204 where the ones that where included in the task. but it seems it was still able to pick up some extra duplicates.~~

~~in total,  there are 1128830 statuses  that have a facility~~

~~for the LAWTON CASELOAD (BECKHAM-CUSTER)  facility alone there are 1135 status with this facility, and only 3 of them have an id~~

~~should I expand my rake task to not focus on the distinct values  and open it up to the duplicates?~~

EDIT: Decided to apply rake tasks to all statuses


lib/tasks/doc_facility.rake
```diff
@@ -0,0 +1,56 @@
...
+    
+
+
+    CSV.foreach('lib/Prison.Facilities-Table.1.csv') do |row|
+        
+        next if row[0] == 'FACILITY'
+        name= row[0]
+        
+        csv << name
+      end
```
sgelbart, line 23: I think your indentation is off

lib/tasks/doc_facility.rake
```diff
@@ -0,0 +1,56 @@
...
+
+namespace :doc do
+  desc 'fill out doc_facility table and link records to status and profile'
+  task :doc_facilities do 
+    csv_path = 'lib/Prison.Facilities-Table.1.csv'
+    csv = []
+    facil=Doc::Status.distinct.pluck(:facility)
+    facil_prof=Doc::Profile.distinct.pluck(:facility)
+    status_profile_facil = facil + facil_prof
+    status_profile_facil = status_profile_facil.uniq
```
sgelbart, line 12: not blocking but you should be able to one-line this with (facil + facil_prof).uniq

lib/tasks/doc_facility.rake
```diff
@@ -0,0 +1,56 @@
...
+
+    
+
+
+    CSV.foreach('lib/Prison.Facilities-Table.1.csv') do |row|
+        
+        next if row[0] == 'FACILITY'
```
sgelbart, line 19:
```suggestion
    CSV.foreach('lib/Prison.Facilities-Table.1.csv', headers: true)) do |row|
```
^^ if you add headers:true it should tell it to skip the first line so you don't need to do `next if row[0] == 'FACILITY'`

lib/tasks/doc_facility.rake
```diff
@@ -0,0 +1,56 @@
+require 'open-uri'
+require 'csv'
+
+namespace :doc do
+  desc 'fill out doc_facility table and link records to status and profile'
+  task :doc_facilities do 
+    csv_path = 'lib/Prison.Facilities-Table.1.csv'
+    csv = []
```
sgelbart, line 8: maybe we can update csv to be prison_names or something more descriptitive? When I see csv I think it's the literal entire csv rows

lib/tasks/doc_facility.rake
```diff
@@ -0,0 +1,56 @@
...
+        next if row[0] == 'FACILITY'
+        name= row[0]
+        
+        csv << name
+      end
+    
+
+      status_profile_facil.each_with_index do |fac,index| 
+        prison = false
+         prison = csv.include? fac
```
sgelbart, line 28:
```suggestion
         prison = csv.include? fac
```
You don't need to default this to false you only need this line

lib/tasks/doc_facility.rake
```diff
@@ -0,0 +1,56 @@
...
+        
+         status = Doc::Status.where(facility: fac).in_batches 
+         
+           status.update_all(doc_facility_id: facility.id)
+
+        
+         
+
+         
+         profile = Doc::Profile.where(facility: fac).in_batches
```
sgelbart, line 46: I don't think in_batches works here and it actually might not be needed since we're just doing update_all
(I'd use in_batches when we were calling individual updates)

spec/services/importers/party_data_spec.rb
```diff
@@ -4,7 +4,7 @@
   describe '#perform(oscn_id)' do
     it 'saves aliases' do
       path = 'spec/fixtures/party_example.html'
-      html = File.open(path).read
+      html = File.read(path)
```
sgelbart, line 7: did you mean to have this?

AspiringDev51: I think this is a rubocop thing. it seems to prefer  `File.open(path).read` over `File.read(path)` If i remember correctly

db/migrate/20220725152514_status.rb
```diff
@@ -0,0 +1,5 @@
+class Status < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :doc_statuses, :doc_facility, index: true
```
sgelbart, line 3: I think for all `add_reference` it's more important to have a primary_key than an index
Tip: search for `add_reference` in the existing code and check how it was implemented in other places (I do that for almost everything unless I'm 100% sure)

sgelbart:
```suggestion
    add_reference :doc_statuses, :doc_facility, foreign_key: true
```

db/migrate/20220725154404_profile.rb
```diff
@@ -0,0 +1,5 @@
+class Profile < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :doc_profiles, :doc_facility, index: true
```
sgelbart, line 3:
```suggestion
    add_reference :doc_profiles, :doc_facility, foreign_key: true
```

holdenmitchell (approved):
Nice work ✨

PR #89, opened 2022-07-27
 issue 83

AspiringDev51:
**To Test**
run rails console
Rails.application.load_tasks
Rake::Task['doc:sentencing_counties'].invoke


lib/tasks/one-off/doc_sentencing_counties.rake
```diff
@@ -0,0 +1,38 @@
+require 'open-uri'
+
+namespace :doc do
+  desc 'Pull in attorneys from OK Bar site'
```
sgelbart, line 4: I think this is the wrong description

lib/tasks/one-off/doc_sentencing_counties.rake
```diff
@@ -0,0 +1,38 @@
...
+    
+    
+ 
+    sent_counties.each do |sent|
+        
+        
+
+        sentence = Doc::Sentence.where(sentencing_county: sent).in_batches
+        
+        if county_court.include?  sent  and sent != 'UNKNOWN COUNTY COURT'
```
sgelbart, line 19: why not do:
```suggestion
        if sent.uppercase.include? 'COUNTY COURT' and sent != 'UNKNOWN COUNTY COURT'
```
I think then you can remove line 6-8?

spec/factories/doc_sentencing_counties.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :doc_sentencing_county do
+    name { 'MyString' }
```
sgelbart, line 3: are we using faker?

spec/services/importers/party_data_spec.rb
```diff
@@ -4,7 +4,7 @@
   describe '#perform(oscn_id)' do
     it 'saves aliases' do
       path = 'spec/fixtures/party_example.html'
-      html = File.open(path).read
+      html = File.read(path)
```
sgelbart, line 7: did you mean to commit this?

lib/tasks/one-off/doc_sentencing_counties.rake
```diff
@@ -0,0 +1,38 @@
+require 'open-uri'
+
+namespace :doc do
+  desc 'Pull in attorneys from OK Bar site'
+  task :sentencing_counties do
+    county_court = ActiveRecord::Base.connection.execute("SELECT DISTINCT sentencing_county from doc_sentences where sentencing_county ilike '%county court%' order by sentencing_county")
+
+    county_court = county_court.column_values(0).sort
+    sent_counties = ::Doc::Sentence.distinct.pluck(:sentencing_county).sort
```
sgelbart, line 9:
```suggestion
    sent_counties = ::Doc::Sentence.select(:sentencing_county).distinct
```
I'd also consider renaming  sent_counties to `sentencing_county_names` (it's longer but clearer what you're working with...if it's an object vs a string)

sgelbart, line 9: I'd consider renaming  sent_counties to `sentencing_county_names` (it's longer but clearer what you're working with...if it's an object vs a string)
and also in your loop renaming to `sentencing_county_name`

db/migrate/20220728193602_fix_doc_sentencing_counties_id.rb
```diff
@@ -0,0 +1,6 @@
+class FixDocSentencingCountiesId < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :doc_sentences, :doc_sentencing_counties_id, :doc_sentencing_county_id
```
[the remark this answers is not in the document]
AspiringDev51: I'm a little confused, are these different migration files wrong or bad? I thought It was good to have different migration files instead of one big one? plus If I remember correctly  this migration file is new to the project since its dealing with a new table. are you saying I should revert all my migrations that I've done to fit it into one file?

spec/services/importers/party_data_spec.rb
AspiringDev51: Again i think this is a rubocop thing that auto corrects whenever I run it. it seems to prefer the `File.read(path)` over  'File.open(path).read' .

db/migrate/20220728193602_fix_doc_sentencing_counties_id.rb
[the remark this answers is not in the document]
sgelbart: I'd have one migration file per commit per table you're working on

sgelbart (approved):

db/migrate/20220728193602_fix_doc_sentencing_counties_id.rb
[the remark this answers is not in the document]
sgelbart: Not a big deal though, just something nice for future to keep migrations a bit cleaner

db/migrate/20220726202412_create_doc_sentencing_counties.rb
```diff
@@ -0,0 +1,10 @@
+class CreateDocSentencingCounties < ActiveRecord::Migration[6.0]
+  def change
+    create_table :doc_sentencing_counties do |t|
+      t.string :name
```
holdenmitchell, line 4: I think it makes sense to match the backend constraint to the model constraint

```suggestion
      t.string :name, null: false
```

db/migrate/20220728193602_fix_doc_sentencing_counties_id.rb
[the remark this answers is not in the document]
holdenmitchell: ☝️ I agree with @sgelbart. We can just rollback and make this change on the original migration. We can pair on it on Friday

db/migrate/20220729190628_remove_null_restriction.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveNullRestriction < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :doc_sentencing_counties, :county_id, true
```
holdenmitchell, line 3: Same as comment above. Let's roll these into the original migration

db/migrate/20220729201159_add_null_restriction.rb
```diff
@@ -0,0 +1,5 @@
+class AddNullRestriction < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :doc_sentencing_counties, :name, false
```
holdenmitchell, line 3: Same as comment above. Let's roll these into the original migration

spec/models/doc_sentencing_county_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe DocSentencingCounty, type: :model do
+  describe 'validations' do
```
holdenmitchell, line 4: Don't forget to add in the association check for `belongs_to :county, optional: true`

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_05_03_152707) do
+ActiveRecord::Schema.define(version: 2022_07_29_201159) do
```
holdenmitchell, line 13: Lot's of unrelated schema changes. We can fix these when we pair

lib/tasks/one-off/doc_sentencing_counties.rake
```diff
@@ -0,0 +1,23 @@
+require 'open-uri'
+
+namespace :doc do
+  desc 'Populates DocSentencingCounty and assignes it to Doc::Sentence'
+  task :sentencing_counties do
+    sent_counties.each do |sent|
```
holdenmitchell, line 6: Having an issue with this line. Where does the sent_counties come from?

holdenmitchell (changes requested):
I think we can go ahead and make some updates to the migrations as @sgelbart suggested. Also, there are some schema changes that we should resolve.

When I run the rake task, I get the following error. I commented on the problem link in the `doc_sentencing_counties.rake` file

```
NameError: undefined local variable or method `sent_counties' for main:Object
```

db/migrate/20220728193602_fix_doc_sentencing_counties_id.rb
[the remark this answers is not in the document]
AspiringDev51: Sounds good!

db/migrate/20220726202412_create_doc_sentencing_counties.rb
AspiringDev51: should I rollback and edit  the file here or make a new db:migrate? either way I assume we will pair on this on friday correct?

AspiringDev51: Also Fixed the error  by adding sent_counties back in,  I think I accidentally deleted it beforehand.

lib/tasks/one-off/doc_sentencing_counties.rake
AspiringDev51: I believe sent_counties was accidentally deleted. added it back in

spec/models/doc_sentencing_county_spec.rb
AspiringDev51: added in!

holdenmitchell (approved):

PR #90, opened 2022-07-28
ARNALL 221 - Exclude Court Cases Filed in Error

sgelbart:
- Flag court cases as error on import
 - Create rake task to mark as error for previously created
    - Run using: `bundle exec rake update:is_error` 
 - Court cases considered an error if:
   - All their relevant associations are empty
   - Or Their dockets have the text 'CASE FILED IN ERROR'


app/models/court_case.rb
```diff
@@ -33,4 +33,9 @@ class CourtCase < ApplicationRecord
   scope :for_county_name, ->(name) { joins(:county).where(counties: { name: name }) }
 
   delegate :html, to: :case_html
+
+  def check_is_error
+    associations = [parties, current_judge, counsels, counts, events, docket_events]
+    associations.all?(&:blank?) || docket_events.any?(&:error?)
```
sgelbart, line 39: does this list of associations look correct?

app/services/importers/court_case.rb
```diff
@@ -21,19 +21,19 @@ def perform
...
       ActiveRecord::Base.transaction do
-        save_case
         save_parties
         save_judges
         save_attorneys
         save_counts
         save_events
         save_docket_events
         logs.update_logs
+        save_case
```
sgelbart, line 31: this appears to be order in-sensitive

lib/tasks/update.rake
```diff
@@ -78,4 +78,16 @@ namespace :update do
...
+  task is_error: [:environment] do
+    associations = [:parties, :current_judge, :counsels, :counts, :events, :docket_events]
+    bar = ProgressBar.new(CourtCase.all.count)
+    CourtCase.in_batches(of: 1000) do |court_cases|
+      court_cases.includes(associations).each do |court_case|
+        court_case.update!(is_error: true) if court_case.error?
+      end
+      bar.increment!(1000)
+    end
+  end
```
sgelbart, line 92: I put this here instead of as a one-off in case we need to re-run because error requirements change? 
Also there's a couple ways we could auto run these on deploy as part of migrations/data-migrations for full automation

sgelbart: Running this currently takes about 18 minutes locally and there is a similar count on production

spec/models/court_case_spec.rb
```diff
@@ -143,4 +143,24 @@
 
     end
   end
+
+  describe '#error?' do
+    subject { court_case.error?}
+    context 'when there are any associations present' do
+      let( :court_case ) { build(:court_case, docket_events: [build(:docket_event)]) }
+      it { should be false }
+      context 'when there is an error a docket event' do
```
sgelbart, line 152:
```suggestion
      context 'when there is an error on a docket event' do
```

holdenmitchell (approved):
Looks great 👍

PR #96, opened 2022-08-04
Gitignore fix

sgelbart:
Fix to exclude ignored files that had previously been added and also exclude all .keep and ide files
Fix was:
`git rm -rf --cached .`
`git add .`


holdenmitchell (approved):
🏆

PR #97, opened 2022-08-08
Throttle api requests to OSCN

sgelbart:
Limits requests to OSCN

### DIFF
--- a/config/initializers/slowweb.rb
+++ b/config/initializers/slowweb.rb
@@ -1,0 +1,4 @@
+require 'slowweb'
+# todo: there is a bug where this is only called half as many times
+limit = ENV.fetch('OSCN_REQUESTS_PER_MINUTE', 6)*2
+SlowWeb.limit('www.oscn.net', limit, 60)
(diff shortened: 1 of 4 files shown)


config/initializers/slowweb.rb
```diff
@@ -0,0 +1,4 @@
+require 'slowweb'
+# todo: there is a bug where this is only called half as many times
+limit = ENV.fetch('OSCN_REQUESTS_PER_MINUTE', 6)*2
+SlowWeb.limit('www.oscn.net', limit, 60)
```
sgelbart, line 4: We alternatively can use a sidekiq throttling gem here
That one won't do any request limiting other than by limiting workers but if we are 100% sure that we will always only have one request per worker then that's fine.
I think if we want to do throttling long term the best solution is to update this gem to work with sidekiq then we will have the best of both worlds (request limiting that supports sidekiq)

holdenmitchell (approved):
We can continue the discussion around Sidekiq Throttling gem vs. the SlowWeb, but this works as a quick solution. 👍

PR #98, opened 2022-08-08
ARNALL 228 - Fuzzy match by name when linking cases and sentences

sgelbart:
This uses the error logic from #90


sgelbart:
Closing until #90 is merged

PR #100, opened 2022-08-09
deleted 2 migration files and added 1 that renamed id field for doc:s…

AspiringDev51:
…entences and gave the county_id field the ability to be null in sentencing_counties


sgelbart:
We can't actually delete those migrations (since they've already been deployed). If you already have the columns for them in your database what you have to do is tell rails that the migration has already been run by adding a row with the migration number to the migrations table (that's what it uses to track what's been run and what hasn't).

sgelbart (approved):

PR #102, opened 2022-08-10
Add sidekiq throttle and fix url parmeters bug

sgelbart:


config/initializers/sidekiq.rb
```diff
@@ -10,8 +10,8 @@
 
 # Create oscn throttling strategy
 Sidekiq::Throttled::Registry.add(:oscn, {
-  threshold: { limit: ENV.fetch('OSCN_THROTTLE', 240).to_i, period: 1.minute },
-  concurrency: { limit: ENV.fetch('OSCN_CONCURRENCY', 10).to_i }
+  threshold: { limit: ENV.fetch('OSCN_THROTTLE', 6).to_i, period: 1.minute },
+  concurrency: { limit: ENV.fetch('OSCN_CONCURRENCY', 1).to_i }
```
sgelbart, line 14: I noticed this had actually already been implemented. We actually want both this and slowweb until we can update slowweb. Slowweb works for things that aren't in workers but we'll have to use this for workers for now.

app/services/importers/new_court_case.rb
```diff
@@ -8,8 +8,7 @@ def initialize(link)
       @case_number = link.text
       @case_types = CaseType.oscn_id_mapping
       @counties = County.pluck(:name, :id).to_h
-      @uri = URI(link['href'])
-      @params = CGI.parse(uri.query)
+      @params = CGI.parse(URI(link['href']).query).map { |k,v| [k.downcase, v.first] }.to_h
```
sgelbart, line 11: There was a bug where casemasterID could be in the URL as camelcase or all lowercase.

app/services/importers/new_court_case.rb
```diff
@@ -24,11 +23,11 @@ def perform
     private
 
     def oscn_id
-      params['casemasterID'].first.to_i
+      params.fetch("casemasterid")
```
sgelbart, line 26: using `fetch` throws an exception if the value isn't there

holdenmitchell (approved):
👍

Issue #86, opened 2022-07-22
Current Judge Not Linking


holdenmitchell:
# Description

Judges are pulled from the [dropdown on this report](https://www.oscn.net/applications/oscn/report.asp?report=WebJudicialDocketJudgeAll) and are displayed on this screen on the [case page](https://www.oscn.net/dockets/GetCaseInformation.aspx?db=tulsa&number=CF-2020-4861)

![image](https://user-images.githubusercontent.com/8531302/180511662-6b9fd45b-5661-452a-b3f0-8e9a2d587f36.png)

The `court_cases.current_judge_id` should link to the current judge listed on the case, but many of the values are `null`.

Files: `app/services/importers/judge.rb`

Find a court case that is missing a judge:

```ruby
CourtCase.where(current_judge_id: nil)
```

sgelbart:
given: There is a recently created case
when: I view the case
then: I see a judge ID on the new cases that matches what's on the OSCN page

sgelbart:
QA done with help of Josh

sgelbart:
QA done with help of @AspiringDev51 
Tested on OSCN-pd database: court_case id 98345 and 308397

sgelbart:
Currently unassigned judges are coming in with null values
https://www.oscn.net/dockets/GetCaseInformation.aspx?db=tulsa&cmid=3020684
This may not be an issue but we might want to create an "Unassigned" judge so we can know that this data is scraping properly

sgelbart:
No judges were scraped since deployment for oklahoma county. Leaving open and debugging.

sgelbart:
Ran `CourtCaseWorker.new.perform(55, 'CJ-2022-4782', true)` on okc db to test for 994831 - confirmed working.
Error most likely due to data not being present the day reported. Confirm tomorrow.

PR #104, opened 2022-08-15
PD2-157 - Scrape judge before updating cases

sgelbart:
This will ensure we have data recent for all judges before running update. 
Alternatively can do on a cron job but this makes it a bit less error prone in-case running manually.
Also fixes bug where judges were being imported with a trailing comma


holdenmitchell (approved):
👍  I think that this should close #86

PR #106, opened 2022-08-25
hooking up to issue_21 in oscn_scraper

AspiringDev51:


sgelbart (approved):

app/services/importers/new_court_case.rb
```diff
@@ -8,7 +8,7 @@ def initialize(link)
       @case_number = link.text
       @case_types = CaseType.oscn_id_mapping
       @counties = County.pluck(:name, :id).to_h
-      @params = CGI.parse(URI(link['href']).query).map { |k, v| [k.downcase, v.first] }.to_h
+      @params = CGI.parse(URI(link['href']).query).to_h { |k, v| [k.downcase, v.first] }
```
sgelbart, line 11: are you sure this works?

AspiringDev51: I think this was a rubocop change, I don't remember doing this .

sgelbart (dismissed):

spec/fixtures/captcha.html
```diff
@@ -0,0 +1,30 @@
...
+    <form action="/recaptcha/recaptcha.aspx" method="POST">
+      <div class="g-recaptcha" data-sitekey="6Ldu8X0UAAAAAMhAy59I5kHOnLqZ6xlCipROyOZE"></div>
+      <br/>
+      <input type="submit" value="Submit">
+	  <input type='hidden' id='source_uri' name='source_uri' value='L2RvY2tldHMvR2V0Q2FzZUluZm9ybWF0aW9uLmFzcHg/ZGI9dHVsc2EmbnVtYmVyPVRSLTIwMjItNzQzOQ=='>
+	  </script>
+    </form>
+	</div>
+  </body>
+</html>
```
aripickard, line 30: Oooo what is this doing? Still getting familiar with the framework.

spec/rails_helper.rb
```diff
@@ -1,6 +1,8 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require 'simplecov'
 require 'simplecov-json'
+require 'webmock/rspec'
+WebMock.disable_net_connect!(allow_localhost: true)
```
aripickard, line 5: What does this do?

db/schema.rb
```diff
@@ -124,6 +124,7 @@
     t.datetime "updated_at", precision: 6, null: false
     t.jsonb "logs"
     t.bigint "current_judge_id"
+    t.boolean "is_error", default: false, null: false
```
aripickard, line 127: What's this for?

aripickard:
I think I need some context on what you're trying to accomplish with these changes, until I get more familiar with this project!

spec/rails_helper.rb
sgelbart: This prevents actually making a request to the real oscn website while we're testing
It forces us to mock up the responses or it throws an erroe

db/schema.rb
sgelbart: Will let it go as a non-blocking fix this time but shouldn't be any changes here if no migration files present.

spec/fixtures/captcha.html
aripickard: That's awesome thank you!

aripickard (approved):
Looks good! :)

PR #107, opened 2022-08-25
Proxy Configuration

sgelbart:
* Enable the proxy server
* Fix to allow for more customizable date ranges for recent cases


app/services/scrapers/recent_cases.rb
```diff
@@ -56,7 +56,7 @@ def fetch_data(county, date, case_type_oscn_id)
     end
 
     def date_range
-      (days_ago.days.ago.to_date..days_forward.days.from_now.to_date).to_a
+      (Date.current - days_ago..Date.current + days_forward).to_a
```
sgelbart, line 59: day.ago and days.from_now don't let us use negative values
Implementing this allows us to use negative values to pull from only future days or only past days that don't go up to the current day

holdenmitchell (approved):
👍 Nice that HTTP Party has a config for this!

PR #110, opened 2022-08-26
fix slowweb

sgelbart:


config/initializers/slowweb.rb
```diff
@@ -1,4 +1,4 @@
 require 'slowweb'
 # todo: there is a bug where this is only called half as many times
-limit = ENV.fetch('OSCN_REQUESTS_PER_MINUTE', 6)*2
+limit = ENV.fetch('OSCN_REQUESTS_PER_MINUTE', 12).to_i*2
```
sgelbart, line 3: slow web globally throttles all requests to specified APIs
This code checks for an environment variable named OSCN_REQUESTS_PER_MINUTE and defaults it to 12 if not provided  
I added to .to_i to force the variable to be an integer because environment variables actually always come in as strings which was causing an issue.

aripickard: Fabulous 👍

aripickard (approved):

config/initializers/slowweb.rb
AspiringDev51, line 3: it seems that here you change  the request per minute to be 12 instead of 6 and made sure to convert this environment variable to an integer before multiplying it by 2. interesting!

AspiringDev51 (approved):
Looks good!

PR #111, opened 2022-08-29
Jw issue 109

AspiringDev51:
Used Rails Migration to add the PDF,Booking and Offense tables for okc_blotter


sgelbart (commented):
Looks good overall! 
A few things:
* let's actually namespace the models (the same way we do the DOC models)? That'll keep it a bit tidier since we'll have tons of models for our different data sources
* can you also add associations to the models and validate them? (let me know if you need help with this)
*  instead of integers, can you do foreign keys on the tables for pdf_id and booking_id (those are for the associations) ` t.references :pdf_id, index: true, foreign_key: true` (I think is the syntax). Remember to rollback your migrations instead of adding new ones
*  for now you actually remove the person_id field from bookings (or at least make it optional)...not sure if we're going to use it or not

app/services/importers/new_court_case.rb
```diff
@@ -8,7 +8,7 @@ def initialize(link)
       @case_number = link.text
       @case_types = CaseType.oscn_id_mapping
       @counties = County.pluck(:name, :id).to_h
-      @params = CGI.parse(URI(link['href']).query).map { |k, v| [k.downcase, v.first] }.to_h
+      @params = CGI.parse(URI(link['href']).query).to_h { |k, v| [k.downcase, v.first] }
```
sgelbart, line 11: I believe rubocop has been auto-correcting this and somehow it keeps getting switched back...

app/models/OkcBlotter/booking.rb
```diff
@@ -0,0 +1,4 @@
+class OkcBlotter::Booking < ApplicationRecord
```
sgelbart, line 1: This inherits from ApplicationRecord which means it's a model
Models in rails allow you to interact with a database table without using SQL and should contain information about their associations, validation, etc.

app/models/OkcBlotter/booking.rb
```diff
@@ -0,0 +1,4 @@
...
+  belongs_to :offense, class_name: 'OkcBlotter::Offense', foreign_key: 'offense_id'
+  validates  :transient, :inmate_number, :booking_number, :booking_date, presence: true
```
sgelbart, line 3: The namespace looks good!
You have the associations backwards though...
PDF has many Bookings has many Offenses
Offenses belongs to Booking belongs to PDF
so the bookings table has pdf_id
and the offenses table has booking_id

db/migrate/20220826221712_create_okc_blotter_bookings.rb
```diff
@@ -0,0 +1,21 @@
+class CreateOkcBlotterBookings < ActiveRecord::Migration[6.0]
+  def change
+    create_table :bookings do |t|
+      t.references :pdfs, null: false, foreign_key: true
```
sgelbart, line 4: this is correct

db/migrate/20220826221907_create_okc_blotter_offenses.rb
```diff
@@ -0,0 +1,16 @@
+class CreateOkcBlotterOffenses < ActiveRecord::Migration[6.0]
+  def change
+    create_table :offenses do |t|
+      t.references :bookings, null: false, foreign_key: true
```
sgelbart, line 4: this is correct

db/migrate/20220831204508_add_bookings_to_pdf.rb
```diff
@@ -0,0 +1,9 @@
+class AddBookingsToPdf < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :pdfs, :booking, null: false, foreign_key: true
+    add_reference :bookings, :offense, null: false, foreign_key: true
```
sgelbart, line 4: I think you thought you needed this because you had the belongs_to/has_many mixed up but once you fix that you can remove

PR #118, opened 2022-09-20
Tulsa Blotter model,factories,and specs

AspiringDev51:
to Test: `bundle exec rspec .`

this PR holds the models ,specs, factories and migration files for the tulsa blotter tables

### DIFF
--- a/app/models/TulsaBlotter/arrest.rb
+++ b/app/models/TulsaBlotter/arrest.rb
@@ -1,5 +1,5 @@
 class TulsaBlotter::Arrest < ApplicationRecord
-  belongs_to :inmate, foreign_key: 'tulsa_blotter_inmates_id', optional: true
+  belongs_to :inmate, foreign_key: 'tulsa_blotter_inmates_id'
   has_many :offenses, class_name: 'TulsaBlotter::Offense', foreign_key: 'tulsa_blotter_arrests_id'
   validates :arrest_date, :arrest_time, :arrested_by, presence: true
 end
--- a/app/models/TulsaBlotter/arrest.rb
+++ b/app/models/TulsaBlotter/arrest.rb
@@ -1,5 +1,5 @@
 class TulsaBlotter::Arrest < ApplicationRecord
   belongs_to :inmate, optional: true
-  has_many :tulsa_offenses, foreign_key: 'arrest_id'
+  has_many :offenses,class_name: 'TulsaBlotter::Offense', foreign_key: 'arrest_id'
   validates :arrest_date, :arrest_time, :arrested_by, presence: true
 end
--- a/app/models/TulsaBlotter/arrest.rb
+++ b/app/models/TulsaBlotter/arrest.rb
@@ -1,5 +1,5 @@
 class TulsaBlotter::Arrest < ApplicationRecord
-  belongs_to :inmate, optional: true
-  has_many :offenses,class_name: 'TulsaBlotter::Offense', foreign_key: 'arrest_id'
+  belongs_to :inmate, foreign_key: 'tulsa_blotter_inmates_id', optional: true
+  has_many :offenses, class_name: 'TulsaBlotter::Offense', foreign_key: 'tulsa_blotter_arrests_id'
   validates :arrest_date, :arrest_time, :arrested_by, presence: true
 end
--- a/app/models/TulsaBlotter/arrest.rb
+++ b/app/models/TulsaBlotter/arrest.rb
@@ -1,0 +1,5 @@
+class TulsaBlotter::Arrest < ApplicationRecord
+  belongs_to :inmate, optional: true
+  has_many :tulsa_offenses, foreign_key: 'arrest_id'
+  validates :arrest_date, :arrest_time, :arrested_by, presence: true
+end
--- a/app/models/TulsaBlotter/inmate.rb
+++ b/app/models/TulsaBlotter/inmate.rb
@@ -1,0 +1,5 @@
+class TulsaBlotter::Inmate < ApplicationRecord
+  belongs_to :roster, optional: true
+  has_many :arrests, foreign_key: 'inmate_id'
+  validates :first, :middle, :last, :gender, presence: true
+end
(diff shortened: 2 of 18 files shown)


app/models/TulsaBlotter/arrest.rb
```diff
@@ -0,0 +1,5 @@
+class TulsaBlotter::Arrest < ApplicationRecord
+  belongs_to :inmate, optional: true
```
sgelbart, line 2: Are you sure this doesn't need to have the foreign_key listed since these are namespaces?
Also, I'd say this shouldn't be optional since you need to have an inmate for there to be an arrest

db/migrate/20220919153449_create_tulsa_blotter_inmates.rb
```diff
@@ -0,0 +1,21 @@
+class CreateTulsaBlotterInmates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :inmates do |t|
```
sgelbart, line 3: we'll want to namespace the tables here
so this should be tulsa_blotter_inmates

db/migrate/20220920143924_create_offenses.rb
```diff
@@ -0,0 +1,15 @@
+class CreateOffenses < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tulsa_offenses do |t|
```
sgelbart, line 3: this name should be tulsa_blotter_offenses

sgelbart (changes requested):
The tables need to have tulsa_blotter prefix
Specify the prefix for the whole module the same way they did in 
models/doc.rb

spec/factories/tulsa/arrests.rb
```diff
@@ -0,0 +1,12 @@
+FactoryBot.define do
+  factory :arrest do
+    inmate { nil }
+    arrest_date { Faker::Lorem.word }
```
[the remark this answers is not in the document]
sgelbart: Here is a full list of types https://github.com/faker-ruby/faker#default

spec/factories/tulsa/inmates.rb
```diff
@@ -0,0 +1,17 @@
+FactoryBot.define do
+  factory :inmate do
+    first { Faker::Lorem.word }
+    middle { Faker::Lorem.word }
+    last { Faker::Lorem.word }
+    gender { Faker::Lorem.word }
+    roster { '' }
```
sgelbart, line 7: if you leave this off completely I think it auto generates the assocation
(so just remove the {  '' }

app/models/TulsaBlotter/offense.rb
```diff
@@ -0,0 +1,4 @@
+class TulsaBlotter::Offense < ApplicationRecord
+  belongs_to :arrest, optional: true
```
sgelbart, line 2: this also isn't optional

sgelbart (approved):
nice job!

app/models/TulsaBlotter/inmate.rb
```diff
@@ -0,0 +1,5 @@
+class TulsaBlotter::Inmate < ApplicationRecord
+  belongs_to :roster, optional: true
```
aripickard, line 2: Should this be optional?

app/models/TulsaBlotter/inmate.rb
```diff
@@ -0,0 +1,5 @@
...
+  has_many :arrests, class_name: 'TulsaBlotter::Arrest', foreign_key: 'tulsa_blotter_inmates_id'
+  validates :first, :middle, :last, :gender, presence: true
```
aripickard, line 4: Not sure if there will always be a middle name!

app/models/TulsaBlotter/offense.rb
```diff
@@ -0,0 +1,4 @@
+class TulsaBlotter::Offense < ApplicationRecord
+  belongs_to :arrest
+  validates :case_number, :court_date, :bond_type, :disposition, presence: true
```
aripickard, line 3: Pretty sure all of these fields will sometimes be null when a person is arrested. I think the `description` is the only one that should always be present.

db/migrate/20220919153449_create_tulsa_blotter_inmates.rb
```diff
@@ -0,0 +1,21 @@
+class CreateTulsaBlotterInmates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tulsa_blotter_inmates do |t|
+      t.string :first
+      t.string :middle
+      t.string :last
+      t.string :gender
+      t.references :roster,null: true, foreign_key: true
+      t.references :booking,null: true, foreign_key: true
```
aripickard, line 9: what is this `booking` table?

db/migrate/20220919163350_create_arrests.rb
```diff
@@ -0,0 +1,16 @@
+class CreateArrests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tulsa_blotter_arrests do |t|
+      t.references :tulsa_blotter_inmates, null: true, foreign_key: true
+      t.string :arrest_date
```
aripickard, line 5: Do we want to store these date fields as strings or parse them into dates? Not sure what's best

db/migrate/20220920143924_create_offenses.rb
```diff
@@ -0,0 +1,15 @@
+class CreateOffenses < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tulsa_blotter_offenses do |t|
+      t.references :tulsa_blotter_arrests, null: true, foreign_key: true
+      t.string :description
+      t.string :case_number
+      t.string :court_date
+      t.string :bond_type
+      t.string :bound_amount
```
aripickard, line 9: Not sure if there's a reason to store `bond_amount` as a float instead of string? It's usually in $0.00 format on the website.

aripickard:
Oops I couldn't remember if we'd decided that I didn't need to review this and noticed after I left comments that it's already been merged haha... Sorry about that. At least it's good experience for me to look over ruby on rails code for a data model I'm familiar with!

db/migrate/20220919153449_create_tulsa_blotter_inmates.rb
sgelbart: This is a mistake

PR #120, opened 2022-10-04
added where_not clause to exclude parties with no oscn_id  

AspiringDev51:
in low and high priority party scrapers

also removed presence  validation for oscn_id in party's model
To test run rspec  on spec/services/scrapers/parties/high_priority.rb 
and
 spec/services/scrapers/parties/hlow_priority.rb

### DIFF
--- a/spec/services/scrapers/parties/high_priority.rb
+++ b/spec/services/scrapers/parties/high_priority.rb
@@ -13,6 +13,15 @@ RSpec.describe Scrapers::Parties::HighPriority do
         end
       end
     end
+    context 'when there is a party with no html and no oscn_id' do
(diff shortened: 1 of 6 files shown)


spec/services/scrapers/parties/high_priority.rb
```diff
@@ -13,6 +13,15 @@
         end
       end
     end
+    context 'when there is a party with no html and no oscn_id' do
+      let!(:party_no_oscn) { create(:party, oscn_id: nil) }
+
+      it 'doesnt add jobs to the PartyWorker' do
```
sgelbart, line 19: instead of "doesn't" I typically use "does not" (or else I think you I don't have to use double quotes or a \ to escape the single quote)

sgelbart (approved):

PR #121, opened 2022-10-05
Oklahoma City Jail Blotter Import

sgelbart:
Create command to import from the oklahoma city jail blotter


test.js
```diff
@@ -0,0 +1,3 @@
+const okcb = require('okcb')
+const test = 'yolo'
+console.log(test)
```
sgelbart, line 3: remove this

spec/rails_helper.rb
```diff
@@ -75,3 +78,43 @@
     with.library :rails
   end
 end
+
+VCR.configure do |c|
```
sgelbart, line 82: add support for VCRs for recording real interactions

spec/services/importers/okc_blotter/pdf_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+        VCR.use_cassette "scrape_pdf" do
+          expect { Importers::OkcBlotter::Pdf.perform('2022-10-01') }
+            .to change { ::OkcBlotter::Pdf.count }
+            .and change { ::OkcBlotter::Booking.count }
+            .and change { ::OkcBlotter::Offense.count }
+        end
+      end
+    end
+  end
+end
```
sgelbart, line 19: this test is a bit minimalistic but I think it's fine for now.

spec/rails_helper.rb
```diff
@@ -75,3 +78,43 @@
...
+  # noise in our cassettes unless we tell VCR to ignore these requests.
+  c.ignore_hosts "chromedriver.storage.googleapis.com"
+
+  c.preserve_exact_body_bytes do |http_message|
+    http_message.body.encoding.name == 'ASCII-8BIT' ||
+      !http_message.body.valid_encoding?
+  end
+
+  # Private environment variables must be set during test run to generate real data for the VCR
+  # They then are then hidden
```
sgelbart, line 105: Here and bellow I'm adding support for obfuscating environment variables provided in this list (so they don't appear in the "VCR" generated)

Gemfile
```diff
@@ -40,13 +41,15 @@ group :development, :test do
   gem 'rubocop', '~> 1.12'
   gem 'rubocop-rails', '~> 2.9'
   gem 'rubocop-rspec'
-  gem 'simplecov'
-  gem 'simplecov-json'
```
sgelbart, line 44: moved gems only used for testing to the "test" group (so they only are installed on test environemnts)

Gemfile
```diff
@@ -40,13 +41,15 @@ group :development, :test do
...
 end
 
 group :test do
   gem 'database_cleaner'
   gem 'shoulda-matchers'
+  gem 'webmock'
+  gem 'simplecov'
+  gem 'simplecov-json'
+  gem 'vcr'
 end
```
sgelbart, line 53: added webmock and VCR gem. Together you can make real HTTP requests in your tests and then store them to "play back" (so you can test real request/response flow without actually making the requests)

app/models/OkcBlotter/booking.rb
```diff
@@ -1,6 +1,5 @@
 class OkcBlotter::Booking < ApplicationRecord
-  belongs_to :pdf, class_name: 'OkcBlotter::Pdf', foreign_key: 'pdfs_id'
+  belongs_to :pdf, class_name: 'OkcBlotter::Pdf'
   belongs_to :roster, optional: true
-  has_many :offense, dependent: :destroy, foreign_key: 'bookings_id'
-  validates :transient, :inmate_number, :booking_number, :booking_date, presence: true
+  has_many :offenses, class_name: 'OkcBlotter::Offense', dependent: :destroy
```
sgelbart, line 4: - removed specification of foreign key (didn't need since it follows the convention)
- removed validates for now until we know exactly what's coming through and what's not

app/models/OkcBlotter/offense.rb
```diff
@@ -1,4 +1,8 @@
 class OkcBlotter::Offense < ApplicationRecord
-  belongs_to :booking, class_name: 'OkcBlotter::Booking', foreign_key: 'bookings_id'
-  validates :type, :charge, presence: true
+  belongs_to :booking, class_name: 'OkcBlotter::Booking'
+
+  class ::ActiveRecord::Base
+    # disable STI to allow columns named "type"
+    self.inheritance_column = :_type_disabled
+  end
```
sgelbart, line 7: we have a field called "type" which is used in rails for single table inheritance so we have to disable that feature to keep the column name

app/models/OkcBlotter/pdf.rb
```diff
@@ -1,3 +1,3 @@
 class OkcBlotter::Pdf < ApplicationRecord
-  has_many :booking, dependent: :destroy, foreign_key: 'pdfs_id'
+  has_many :bookings, dependent: :destroy
```
sgelbart, line 2: has_many should be plural

app/models/okc_blotter.rb
```diff
@@ -0,0 +1,5 @@
+module OkcBlotter
+  def self.table_name_prefix
+    'okc_blotter_'
+  end
```
sgelbart, line 4: specifies the table prefix for all of the models inside this module
(so the table for OkcBlotter::Booking becomes okc_blotter_bookings)

app/services/application_service.rb
```diff
@@ -0,0 +1,5 @@
+class ApplicationService
+  def self.perform(*args, &block)
+    new(*args, &block).perform
+  end
+end
```
sgelbart, line 5: This class helps solidify an established pattern we already have in our code...
all of our services have a single entry point (which makes it clear how to use them)
To implement that they each have an object method called `perform` (self.perform) which instantiates the object with all the passed in parameters and then calls the `perform` method on  object itself (which is where all the logic is created).

app/services/bucket.rb
```diff
@@ -5,6 +5,10 @@ def initialize
     @s3 = client
   end
 
+  def put_object(filepath, buffer)
+    s3.put_object(bucket: ENV['BUCKETEER_BUCKET_NAME'], key: filepath, body: buffer)
+  end
```
sgelbart, line 10: store object

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+      end
+
+      def save
+        bookings_hash = json
+        ::OkcBlotter::Pdf.create!(
+          parsed_on: DateTime.now,
+          date: @date,
+          bookings: bookings_hash.map do |booking_hash|
+            build_bookings(booking_hash)
+          end
```
sgelbart, line 36: here we're using `.map` to loop through the bookings hash and call the `build_bookings` method to build an object for each item in the hash 
(map is like each but it creates a new array with the returned results)

db/migrate/20220919153449_create_tulsa_blotter_inmates.rb
```diff
@@ -6,7 +6,6 @@ def change
       t.string :last
       t.string :gender
       t.references :roster,null: true, foreign_key: true
-      t.references :booking,null: true, foreign_key: true
```
sgelbart, line 9: this was a mistake

db/migrate/20221003201309_rename_blotter_tables.rb
```diff
@@ -0,0 +1,14 @@
+class RenameBlotterTables < ActiveRecord::Migration[6.0]
+  def change
+    rename_table :pdfs, :okc_blotter_pdfs
+    rename_table :bookings, :okc_blotter_bookings
+    rename_table :offenses, :okc_blotter_offenses
```
sgelbart, line 5: add namespaces to the tables

db/migrate/20221003201309_rename_blotter_tables.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    rename_column(:okc_blotter_bookings, :pdfs_id, :pdf_id)
+    rename_column(:okc_blotter_offenses, :bookings_id, :booking_id)
```
sgelbart, line 8: make follow the standard

db/migrate/20221003201309_rename_blotter_tables.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    change_column(:okc_blotter_bookings, :booking_date, :datetime)
+    change_column(:okc_blotter_bookings, :release_date, :datetime)
+    change_column(:okc_blotter_offenses, :bond, :decimal, precision: 10, scale: 2)
```
sgelbart, line 12: change format

db/migrate/20221005163229_update_okc_blotter_pdfs.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateOkcBlotterPdfs < ActiveRecord::Migration[6.0]
+  def change
+    change_column(:okc_blotter_pdfs, :parsed_on, :datetime)
```
sgelbart, line 3: we try to merge migration files that act on the same table into one but these migrations are a bit finicky so doing as two

db/schema.rb
```diff
@@ -10,32 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_09_20_143924) do
+ActiveRecord::Schema.define(version: 2022_10_05_163229) do
```
sgelbart, line 13: this file is automatically generated to reflect the migrations to the schema

lib/tasks/okc_blotter.rake
```diff
@@ -0,0 +1,10 @@
+namespace :okc_blotter do
+  desc 'Scape and parse'
+  task import_new: [:environment] do
+    Importers::OkcBlotter::Pdf.import_since_last_run
+  end
+
+  task import_missing: [:environment] do
+    Importers::OkcBlotter::Pdf.import_missing_dates
+  end
+end
```
sgelbart, line 10: this is a Rake task which allows us to call ruby code from the command line

spec/rails_helper.rb
```diff
@@ -1,6 +1,9 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require 'simplecov'
 require 'simplecov-json'
+require 'webmock/rspec'
+
+WebMock.disable_net_connect!
```
sgelbart, line 6: this ensures we don't do any real requests during tests (unless we're generating a "cassette" with VCR

app/services/importers/okc_blotter/pdf.rb
AspiringDev51: Got it!

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+        )
+      end
+
+      def json
```
AspiringDev51, line 40: A little confused on what this does after reading the hack.

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+        def pdf.path
+          '/path/hack.pdf' # this is a hack see https://github.com/jnunemaker/httparty/issues/675#issuecomment-590757288
+        end
+
+        return @json if @json
+
+        response = HTTParty.post(
+          "#{url}/parse",
```
AspiringDev51, line 48: used to connect to the app and post to it, also using the parse parameter. but since this is a post and not a get shouldn't the only response be something like a header like 200? also the body is the unparsed pdf right?

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+        )
+        @json = JSON.parse(response.body)
+        return @json if response.success?
+
+        raise StandardError.new "Error parsing pdf: #{@json['message']}"
+      end
+
+      def pdf
+        return @pdf if @pdf
+        pdf = Bucket.new.get_object("#{s3_path}/#{@date}.pdf").body rescue false
```
AspiringDev51, line 60: where pdf is defined, if null runs the download_pdf_from_website method. pdf found in s3 bucket/folder that matches corresponding date

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      def url
+        'https://okc-blotter.herokuapp.com'
+      end
+
+      def auth_token
+        ENV['OKC_BLOTTER_AUTH_TOKEN']
+      end
+
+      def self.import_since_last_run
```
AspiringDev51, line 108: method to run the perform method from the last import date to today

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+        last_run_date = ::OkcBlotter::Pdf.order(date: :desc).first.date
+        (last_run_date + 1.day..DateTime.now.to_date).each do |date|
+          begin
+            perform(date)
+          rescue StandardError => error
+            puts "error processing #{date}. Data not saved" #todo: log these somewhere?
```
AspiringDev51, line 114: if no recent pdf to grab last_run_date  the catch error?

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+            puts "error was: #{error}"
+          end
+        end
+      end
+
+      # Note: PDFs only are on the website for 30 days.
+      # If the pdfs have already been saved to s3 you may be able to parse longer
+      def self.import_missing_dates(since_date = (DateTime.now - 1.month).to_date)
+        query = <<-sql
```
AspiringDev51, line 123: performs a sql query to find missing dates (since this importer should get a pdf every day?) then  runs the perform method using these dates?

db/migrate/20221003201309_rename_blotter_tables.rb
AspiringDev51: module_class_name , got it

db/migrate/20221005163229_update_okc_blotter_pdfs.rb
AspiringDev51: Understandable

lib/tasks/okc_blotter.rake
AspiringDev51: got it, want to have quick access to your importers functionality.

spec/cassettes/scrape_pdf.yml
```diff
@@ -0,0 +1,1342 @@
+---
```
AspiringDev51, line 1: this seems to be a type of configuration file for mocking scraping pdfs instead of making the requests right?

spec/models/okc/okc_blotter_booking_spec.rb
```diff
@@ -1,15 +1,9 @@
 require 'rails_helper'
```
AspiringDev51, line 2: since you removed the validations you needed to cut them down here as well, got it.

spec/models/okc/okc_blotter_offense_spec.rb
```diff
@@ -1,10 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe OkcBlotter::Offense, type: :model do
-  describe 'validations' do
```
AspiringDev51, line 4: same as booking_spec

spec/models/okc/okc_blotter_pdf_spec.rb
```diff
@@ -2,6 +2,6 @@
 
 RSpec.describe OkcBlotter::Pdf, type: :model do
   describe 'validations' do
-    it { should have_many(:booking).dependent(:destroy) }
+    it { should have_many(:bookings).dependent(:destroy) }
```
AspiringDev51, line 5: have_many relationships are plural. got it.

spec/rails_helper.rb
AspiringDev51: new to VCR, interested to learn more!

spec/rails_helper.rb
AspiringDev51: Basically hiding them from when the VCR takes a snap shot so sensitive info is not included right?

spec/services/importers/okc_blotter/pdf_spec.rb
AspiringDev51: got it, Its easy to follow at least!

AspiringDev51 (commented):
I think I understand what you are doing here basically making the importer for the PDF and to create its associated bookings and offenses. I don't think I should approve now, as I'm still unsure about bugs.

app/models/OkcBlotter/booking.rb
sgelbart: Offenses had the foreign key of booking_id
Really you never need to specify foreign key here unless it's something weird
Here it's offense_id so we're fine. The weird thing in this case is the class name since it's names paced so that's what we have to specify.

app/models/OkcBlotter/offense.rb
sgelbart: yup

app/models/OkcBlotter/pdf.rb
sgelbart: nope

app/services/bucket.rb
```diff
@@ -14,7 +18,7 @@ def get_object(filepath)
   def client
     credentials = Aws::Credentials.new(ENV['BUCKETEER_AWS_ACCESS_KEY_ID'], ENV['BUCKETEER_AWS_SECRET_ACCESS_KEY'])
     Aws.config.update(
-      region: 'us-east-1',
+      region: ENV.fetch('BUCKETEER_AWS_REGION', 'us-east-1'),
```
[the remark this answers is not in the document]
sgelbart: yup

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+
+module Importers
+  module OkcBlotter
+    class Pdf < ApplicationService
+      def initialize(date)
+        @date = date
+      end
+
+      def perform
+        download_pdf_from_website unless pdf
```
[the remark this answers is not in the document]
sgelbart: yup

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,139 @@
...
+      def download_pdf_from_website
+        input = HTTParty.get("#{url}/pdfs?date=#{@date}",
+                             headers: {
+                               Authorization: auth_token
+                             }).body
+        Zip::InputStream.open(StringIO.new(input)) do |io|
+          while entry = io.get_next_entry #note: there is only actually one pdf here
+            pdf = io.read
+            filename = entry.name
+            Bucket.new.put_object("#{s3_path}/#{filename}", pdf)
```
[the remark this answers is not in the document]
sgelbart: yup

app/services/importers/okc_blotter/pdf.rb
sgelbart: we're POSTing a PDF to the endpoint. We're just storing the PDF in memory but HTTParty only supports it if we're saving the PDF first to the local file system.  There's no need for us to do that since we're already saving it to s3, but HTTPparty is expecting it to have a path parameter indicating that we're passing in a file. This is a hack to get around that.

app/services/importers/okc_blotter/pdf.rb
sgelbart: post requests almost always have bodies too
in traditional CRUD, when doing a create you'd get back the object you created.
There's some grey area about what should be GET or POST but since this a submitting a file it definitely needs to be POST.

spec/rails_helper.rb
sgelbart: yup!

app/services/importers/okc_blotter/pdf.rb
AspiringDev51: Ok, I think I understand, would love to half a walkthrough of this though.

app/services/importers/okc_blotter/pdf.rb
```diff
@@ -0,0 +1,137 @@
...
+        (last_run_date + 1.day..DateTime.now.to_date).each do |date|
+          perform(date)
+        rescue StandardError => e
+          puts "error processing #{date}. Data not saved" # TODO: log these somewhere?
+          puts "error was: #{e}"
+        end
+      end
+
+      # NOTE: PDFs only are on the website for 30 days but if pdfs are saved to s3 you may be able to parse longer
+      def self.import_missing_dates(since_date = (DateTime.now - 1.month).to_date)
```
AspiringDev51, line 122: seems to collect dates and run the perform method on them, but from what I saw the perform method doesn't have a date parameter

db/schema.rb
```diff
@@ -160,33 +159,6 @@
     t.datetime "updated_at", precision: 6, null: false
   end
 
-  create_table "doc_historical_sentences", force: :cascade do |t|
```
AspiringDev51, line 163: seems you deleted this table, was it not useful?

spec/rails_helper.rb
```diff
@@ -101,6 +104,11 @@
   # Private environment variables must be set during test run to generate real data for the VCR
   # They then are then hidden
   env_keys = %w[
+    BUCKETEER_AWS_SECRET_ACCESS_KEY
+    BUCKETEER_BUCKET_NAME
+    OKC_BLOTTER_AUTH_TOKEN
+    OKC_BLOTTER_AUTH_TOKEN
+    BUCKETEER_AWS_ACCESS_KEY_ID
```
AspiringDev51, line 111: adding more environment variable keys?

spec/services/importers/okc_blotter/pdf_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+RSpec.describe Importers::OkcBlotter::Pdf do
+  describe '.perform' do
+    context 'when there is an existing pdf on the website' do
+      it 'saves the data to the database' do
+        # to re-record this test you need to have AWS env variables configured
+        # most of them are obfuscated in the final vcr (see rails_helper.rb)
+        ENV['BUCKETEER_AWS_REGION'] = 'us-west-2'
+        ENV['BUCKETEER_BUCKET_NAME'] = 'bucket-name' # aws s3 changes behavior if bucket is uppercase or has underscores
+        VCR.use_cassette 'scrape_pdf' do
```
AspiringDev51, line 11: is the use of the VCR and cassette new?  I don't remember seeing it last time.

AspiringDev51 (approved):
Not exactly sure what has changed since my last review for a lot of the files,  what stood out was the test  for pdf_spec, which is now using VCR I believe, similar situation as to last time since its hard for me to see any errors, security issues, or places to optimize, but I think the core of the PR hasn't changed and still has the same flow as the last time I reviewed.

PR #123, opened 2022-10-10
Allow user-agent to come from environment variable

sgelbart:


app/services/scrapers/judges.rb
```diff
@@ -17,7 +17,7 @@ def perform
       # rubocop:enable Layout/LineLength
 
       html = HTTParty.get(base_url, headers: {
-                            'User-Agent': user_agent,
+                            'User-Agent': ENV.fetch('USER_AGENT', user_agent),
```
AspiringDev51, line 20: instead of giving  User-Agent a value from a regular variable, you simply gave it a value from an environment variable.My guess as to why is to have the value be tied closer to the  user's session?

config/initializers/proxy.rb
```diff
@@ -1,9 +0,0 @@
-if ENV['PROXY_HOST']
```
AspiringDev51, line 1: interesting, so I'm guessing that the need for a proxy wasn't needed anymore which called for this files deletion right? do the Proxy environment variables still exist as well?

AspiringDev51 (approved):
Easy to follow change, looks good to me! Seems we aren't using it anymore but  I may look into  about what this proxy is in this context to learn a bit more about it.

PR #125, opened 2022-10-12
rake task may be fixed

AspiringDev51:
needed to remove validation for fields in  offenses


db/migrate/20221012164810_create_tulsa_offenses.rb
```diff
@@ -2,11 +2,11 @@ class CreateTulsaOffenses < ActiveRecord::Migration[6.0]
...
-      t.string :court_date
-      t.string :bond_type
-      t.string :bound_amount
-      t.string :disposition
+      t.string :case_number, null: true
+      t.string :court_date, null: true
+      t.string :bond_type, null: true
+      t.string :bound_amount, null: true
+      t.string :disposition, null: true
       t.timestamps
```
sgelbart, line 10: since this is on a migration that has already been deployed this is going to need to be a new migration

PR #126, opened 2022-10-13
Name splitter fix

sgelbart:


app/services/matchers/party_name_splitter.rb
```diff
@@ -3,10 +3,9 @@ module Matchers
   class PartyNameSplitter
     attr_reader :party, :full_name, :split_name, :split_count
 
-    def initialize(party, full_name)
+    def initialize(party)
```
AspiringDev51, line 6: removing the full_name parameter since party should already have a full_name

app/services/matchers/party_name_splitter.rb
```diff
@@ -3,10 +3,9 @@ module Matchers
...
       @party = party
-      @full_name = full_name
-      @split_name = full_name.split(/\p{Space}\p{Space}/)
+      @split_name = party.full_name.squish.split(' ')
```
AspiringDev51: looks like regex which isn't fun, splitting double spaces like how you did looks way better

app/services/matchers/party_name_splitter.rb
```diff
@@ -49,7 +48,7 @@ def update_two_name(split_name)
     end
 
     def clean_string(string)
-      string.gsub(',', '').squish
```
AspiringDev51, line 52: .squish not needed? is it  detrimental using it when its not needed or does it just do nothing?

lib/tasks/update.rake
```diff
@@ -25,7 +25,7 @@ namespace :update do
     # TODO: Move to party importer task
     parties.each do |p|
       bar.increment!
-      Matchers::PartyNameSplitter.new(p, p.full_name).perform
```
AspiringDev51, line 28: matching the parameter change

AspiringDev51 (approved):
looks good to me!

app/services/matchers/party_name_splitter.rb
sgelbart: yup

app/services/matchers/party_name_splitter.rb
sgelbart: it was moved to line 8

PR #128, opened 2022-10-20
Set Party Scraping Priority

sgelbart:
Currently the party scraper is getting push to the default queue but it should be finer tuned than that so we ensure everything is getting processed.


AspiringDev51 (approved):
It seems that for both high and low priority you simply set their queue priority to be either high or low then you ran perform_async. looks good to me!

PR #131, opened 2022-10-28
Add Rake task for OCSO warrants

sgelbart:


lib/tasks/ocso.rake
```diff
@@ -0,0 +1,6 @@
+namespace :ocso do
+  desc 'Scrape ocso warrant data'
+  task daily_import: [:environment] do
+    Scrapers::OcsoWarrants.perform
+  end
+end
```
sgelbart, line 6: This creates a command for us to run the code like `rake ocso:daily_import`
(we'll be calling this from a cron job)

jacobtice (approved):

aripickard (approved):

PR #136, opened 2022-12-19
Parcel for Rails

AspiringDev51:


spec/factories/parcels.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :parcel do
+    geoid20 { 'MyString' }
+    zip { 'MyString' }
+    tract { '' }
+    block { 'MyString' }
+    lat { '9.99' }
+    long { '9.99' }
+  end
```
sgelbart, line 9: can you use Faker for these? Try to make the type match up

db/migrate/20221219171622_create_parcels.rb
```diff
@@ -0,0 +1,18 @@
+class CreateParcels < ActiveRecord::Migration[6.0]
+  def change
+    create_table :parcels do |t|
+      t.string :geoid20,null:false
+      t.string :zip 
+      t.integer :tract ,null:false
+      t.string :block ,null:false
+      t.string :lat, null:true
+      t.string :long , null:true
```
sgelbart, line 9: spaces should be after commas
not sure why we don't have rubocop on migrations. Maybe try turning on?
If that doesn't work then at least get a code beautifier in your IDE to catch this stuff?

AspiringDev51: ok will  look into getting a code linter.

PR #141, opened 2022-12-29
Fix indexs for DOC migrations

sgelbart:


db/migrate/20221213171103_doc_indexes.rb
```diff
@@ -4,9 +4,8 @@ def change
     remove_index :doc_statuses, [:doc_profile_id, :doc_facility_id]
     
     add_index :doc_sentences, [:doc_profile_id, :sentence_id] ,unique:true
-    add_index :doc_statuses, [:doc_profile_id, :doc_facility_id], unique:true
```
sgelbart, line 7: this extra index was left on here so the migrations couldn't run
(since the combination of :doc_profile_id and :doc_facility_id aren't unique in the production database)

db/migrate/20221221201733_add_index_to_doc_offenses.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToDocOffenses < ActiveRecord::Migration[6.0]
+  def change
+    add_index :doc_offense_codes, [:statute_code, :description, :is_violent ],unique:true, name: 'doc_offense_codes_index'
```
sgelbart, line 3: This allows us to do upsert for doc_offense_codes too (need the unique index on the combined fields)

jacobtice (approved):

aripickard (approved):

PR #143, opened 2023-01-10
Add Raygun to Rake Tasks and Sidekiq

sgelbart:


AspiringDev51 (approved):
added sidekiq version of raygun to its initializer and added a task initializer, looks good to me!

PR #144, opened 2023-01-11
Additional Raygun Fixes

sgelbart:
Remove binding.pry (not sure how those made it in)
Wrap the rake override in a begin rescue block (it looks like it doesn't work when being called outside rake)
Better error handling where we had rescues

### DIFF
--- a/app/services/importers/okc_blotter/pdf.rb
+++ b/app/services/importers/okc_blotter/pdf.rb
@@ -113,8 +113,9 @@ module Importers
         (last_run_date + 1.day..DateTime.now.to_date).each do |date|
           perform(date)
         rescue StandardError => e
-          puts "error processing #{date}. Data not saved" # TODO: log these somewhere?
-          puts "error was: #{e}"
+          error = StandardError.new("Error processing #{date}. Data not saved. Message was: #{e}")
+          puts error
+          Raygun.track_exception(error)
         end
       end
 
@@ -128,8 +129,9 @@ module Importers
         missing_dates.each do |date|
           perform(date)
         rescue StandardError => e
-          puts "error processing #{date}. Data not saved" # TODO: log these somewhere?
-          puts "error was: #{e}"
+          error = StandardError.new("Error processing #{date}. Data not saved. Error was: #{e}")
+          puts error
+          Raygun.track_exception(error)
         end
       end
     end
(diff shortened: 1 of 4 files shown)


app/services/importers/okc_blotter/pdf.rb
```diff
@@ -113,8 +113,9 @@ def self.import_since_last_run
         (last_run_date + 1.day..DateTime.now.to_date).each do |date|
           perform(date)
         rescue StandardError => e
-          puts "error processing #{date}. Data not saved" # TODO: log these somewhere?
-          puts "error was: #{e}"
+          error = StandardError.new("Error processing #{date}. Data not saved. Message was: #{e}")
+          puts error
+          Raygun.track_exception(error)
```
sgelbart, line 118: this prints the error and then sends it to raygun (rather than just printing it to console and getting no notifications)

spec/services/scrapers/report_warrants_spec.rb
```diff
@@ -1,8 +1,9 @@
+
 require 'rails_helper'
 
 RSpec.describe Scrapers::OcsoWarrants do
   describe '#perform' do
-    # this test works however it requires a cassett to large to commit
+    # this test works however it requires a cassette too large to commit
```
sgelbart, line 6: random comment fix

config/initializers/task.rb
```diff
@@ -1,18 +1,18 @@
...
+        def execute(args=nil)
+          orig_execute(args)
+        rescue Exception => ex
+          raise ex if [Interrupt, SystemExit, SignalException].include?(ex.class)
+          Raygun.track_exception(ex)
+          raise ex
+        end
       end
     end
+  rescue
```
sgelbart, line 16: remove pry (not sure how rubo didn't catch)
wrap in try/rescue for when we're running outside of a rake task

app/services/importers/okc_blotter/pdf.rb
AspiringDev51: ok! creates a error object then adds it to what I assume is a list in Raygun.

config/initializers/task.rb
AspiringDev51: try catch is begin rescue in ruby right?

AspiringDev51 (approved):
looks good to me!

config/initializers/task.rb
sgelbart: yes being rescue!

PR #148, opened 2023-01-17
tables made for tulsa city, scraper in_progress

AspiringDev51:
### To Test

Open three Command line tabs.

run `redis-server` in one

run ` bundle exec sidekiq` in one

after these are running run `bundle exec rake tulsa_city:inmate_list` in the last tab.  

Inmate record count should equal Inmate_json count( which is used for the progress bar)


config/initializers/sidekiq.rb
```diff
@@ -20,4 +20,13 @@
   }
 })
 
+Sidekiq::Throttled::Registry.add(:tulsa_city, **{
+  threshold: {
+    :limit => ->(*_) {
+      Time.now.utc.between?(OKLAHOMA_DAY_START, OKLAHOMA_DAY_END) ? 60.to_i : 1_000
```
sgelbart, line 26: 60.to_i - pull this from an env variable
ENV.fetch('TULSA_CITY_THROTTLE', 60).to_i

db/migrate/20230117180357_create_tulsa_city_inmates.rb
```diff
@@ -0,0 +1,28 @@
+class CreateTulsaCityInmates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tulsa_city_inmates do |t|
+      t.string :inmateId, index: {unique: true}
```
sgelbart, line 4: good job on unique indexes
Column names are always snake case in ruby e.g., `inmate_id`

sgelbart:
Look like rubocop can run. 
Also in general PRs should be QAed to the acceptance criteria on the ticket (I just went ahead and updated the ACs to make them a bit clearer) https://9bcorp.atlassian.net/browse/DEV-524 - you can include a link in the ticket

app/services/scrapers/TulsaCity/tulsa_inmates.rb
```diff
@@ -0,0 +1,43 @@
+require 'byebug'
```
sgelbart, line 1: not blocking but I don't think you need byebye bug (if you do we should try to move it someplace more global later)

app/services/scrapers/TulsaCity/tulsa_inmates.rb
```diff
@@ -0,0 +1,43 @@
...
+      def self.perform_inmates
+        new.perform_inmates
+      end
+
+      def perform
+        inmate_json = perform_inmates
+        bar = ProgressBar.new(inmate_json.count)
+        inmate_json.each do |inmate|
+          TulsaCityWorker
+            .set(queue: :high)
```
sgelbart, line 27: not blocking but you don't need to set this queue as high here since it's already set line 7 of tulsa_city_worker.rb (you just explicitly set it if you want to use something other than the default)

app/workers/tulsa_city_worker.rb
```diff
@@ -0,0 +1,88 @@
...
+    offense.paragraph = offense_json['paragraph']
+    offense.crime = offense_json['crime']
+
+    offense.inmate_id = inmate.id
+    offense.save!
+    offense
+  end
+
+  def perform_offense(dataid)
+    # 30010
```
sgelbart, line 71: remove comment

app/workers/tulsa_city_worker.rb
```diff
@@ -0,0 +1,88 @@
...
+class TulsaCityWorker
+  include Sidekiq::Worker
+  include Sidekiq::Throttled::Worker
+  sidekiq_options retry: 5, queue: :high
+  sidekiq_throttle_as :tulsa_city
+  def self.perform(recordid, active)
+    new.perform(recordid, active)
+  end
+
+  def perform(incidentrecordid, active)
```
sgelbart, line 13: not blocking but you should snake case this `incident_record_id`

sgelbart: (same with recordid)

app/workers/tulsa_city_worker.rb
```diff
@@ -0,0 +1,88 @@
...
+    inmate.court_division = inmate_json['courtDivision']
+    inmate.incident_record_id = incidentrecordid
+    inmate.dob = inmate_json['DOB']
+    inmate.active = active
+    inmate.save!
+    inmate
+  end
+
+  def save_offense(offense_json, inmate)
+    offense = ::TulsaCity::Offense.find_or_create_by(docket_id: offense_json['docketId'])
```
sgelbart, line 53: This might be okay but I'm actually not 100% sure that multiple inmates couldn't have the same docketId if they were part of the same case
would be safer here to also do it by inmate_id

db/migrate/20230117180357_create_tulsa_city_inmates.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.string :middle_name
+      t.string :last_name
+      t.string :dob
+      t.string :height
+      t.string :weight
+      t.string :hair_color
+      t.string :eye_color
+      t.string :race
+      t.string :gender
+      t.string :arrest_date
```
sgelbart, line 15: This should be a date or a daetetime

db/migrate/20230117180357_create_tulsa_city_inmates.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.string :arresting_officer
+      t.string :arresting_agency
+      t.string :booking_date_time
```
sgelbart, line 18: same

spec/factories/tulsa_city_inmates.rb
```diff
@@ -0,0 +1,22 @@
...
+    eyeColor { 'MyString' }
+    race { 'MyString' }
+    gender { 'MyString' }
+    arrestDate { 'MyString' }
+    arrestingOfficer { 'MyString' }
+    arrestingAgency { 'MyString' }
+    bookingDateTime { 'MyString' }
+    courtDate { 'MyString' }
+    releasedDateTime { 'MyString' }
+    courtDivision { 'MyString' }
```
sgelbart, line 20: Use Faker for these (and use the right type, so for date I think it's `Faker::Date`) 
see https://github.com/faker-ruby/faker

sgelbart (commented):
Looks good! the main things are making sure the dates are the right format
and also that inmate_id is on the find_or_initialize_by for the offense 
Really, we could write some nice tests for this but I want to make sure to give us enough time to QA so we can skip for now and come back and do later if we have time.

app/workers/tulsa_city_worker.rb
sgelbart: `offense = ::TulsaCity::Offense.find_or_create_by(docket_id: offense_json['docketId'], inmate_id: inmate.id)`

app/workers/tulsa_city_worker.rb
```diff
@@ -0,0 +1,88 @@
...
+    offense.case_number = offense_json['caseNumber']
+    offense.court_division = offense_json['courtDivision']
+    offense.hold = offense_json['hold']
+    offense.docket_id = offense_json['docketId']
+    offense.title = offense_json['title']
+    offense.section = offense_json['section']
+    offense.paragraph = offense_json['paragraph']
+    offense.crime = offense_json['crime']
+
+    offense.inmate_id = inmate.id
```
sgelbart, line 65: can remove this

sgelbart (approved):

PR #150, opened 2023-02-28
Add Issues and Issue Party models

holdenmitchell:
# Description

Adds the Issues and Issue Party tables / models for Small Claims cases


holdenmitchell:
Need to de-duplicate parties on Small Claim cases before we can attach parties to the issue

PR #151, opened 2023-03-10
Fix party and counsel bugs

sgelbart:
Fixes 
https://9bcorp.atlassian.net/browse/DEV-757
and 
https://9bcorp.atlassian.net/browse/DEV-689

To clear out extra case party records run the following query:
```sql
delete
from case_parties
where id in (select a.id
             from case_parties a
                      join case_parties b on a.court_case_id = b.court_case_id
                      join parties pa on a.party_id = pa.id
                      join parties pb on b.party_id = pb.id
             where pa.full_name = pb.full_name
               and pa.party_type_id = pb.party_type_id
               and a.id < b.id)
```


holdenmitchell (approved):
👍 LGTM. Is the plan to just execute the SQL query on production once deployed?

Issue #31, opened 2022-01-03
Add additional fields for docket events


holdenmitchell:
# Description

Many of the docket events have standardized text that can be used to extract information. Pull in this standard text as well as any docket event types that are not currently in the database.

PR #152, opened 2023-03-21
Tweak code so that the parser only accepts the link html

holdenmitchell:
# Description

Paired with: [OSCN Scraper 31](https://github.com/AyOK-Code/oscn_scraper/pull/31)

Only pass the link html to the Parser


holdenmitchell:
Need to bump gem version in this PR

PR #156, opened 2023-03-28
changed court_case_worker and high and low priority to use enqueued

AspiringDev51:


app/services/scrapers/high_priority.rb
```diff
@@ -15,10 +15,13 @@ def self.perform(county_name, days_ago: 7, days_forward: 7)
...
     def perform
       cases = fetch_case_list
-
       bar = ProgressBar.new(cases.count)
       puts "#{cases.count} are high priority for update for #{county.name} county"
       cases.each do |case_number|
+        court_case = ::CourtCase.find_by!(county_id: @county.id, case_number: case_number)
+        next unless court_case.enqueued == false
+
+        court_case.update(enqueued: true)
```
sgelbart, line 24: this is one way to solve it but it would be more performant if you switch from doing the database query in maybe the fetch_case_list method to retrieve all of them at once

spec/services/scrapers/low_priority_spec.rb
```diff
@@ -7,9 +7,9 @@
     it 'add specs' do
```
sgelbart, line 7: also rename this to describe what it does :-)
`    it 'enqueues the court case worker' do`

spec/services/scrapers/low_priority_spec.rb
```diff
@@ -4,15 +4,17 @@
...
+    it 'adds jobs to the CourtCaseWorker' do
       test_case = create(:court_case, :with_html, closed_on: 7.months.ago)
       test_case.case_html.update(scraped_at: 7.months.ago)
-
-      data = described_class.perform
+      data = nil
 
       expect do
-        described_class.perform
+        data = described_class.perform
```
sgelbart, line 13: why are you doing this stuff with the data variable?

spec/services/scrapers/low_priority_spec.rb
```diff
@@ -4,15 +4,17 @@
...
       end.to change(CourtCaseWorker.jobs, :size).by(1)
+      expect do
+        described_class.perform
+      end.to change(CourtCaseWorker.jobs, :size).by(0)
```
sgelbart, line 17: in general we only want one assertion per `it` block (so move this to it's own)
you can then move the setup to a `before` block

spec/services/scrapers/low_priority_spec.rb
AspiringDev51: in order for the second test` expect(data.first.id).to eq test_case.id` to  not error out, its making sure the test_case id equals the id of the first case in the job.

spec/services/scrapers/low_priority_spec.rb
```diff
@@ -4,16 +4,29 @@
...
       expect do
         described_class.perform
-      end.to change(CourtCaseWorker.jobs, :size).by(1)
+      end.to change(CourtCaseWorker.jobs, :size).by(0)
+
       expect(data.first.id).to eq test_case.id
     end
+    it 'doesnt add duplicates to worker' do
+      test_case = create(:court_case, :with_html, closed_on: 7.months.ago)
+      test_case.case_html.update(scraped_at: 7.months.ago)
```
sgelbart, line 24: these should be in a `before` block (try to avoid doing setup in the `it` block)

spec/services/scrapers/high_priority_spec.rb
```diff
@@ -4,17 +4,34 @@
...
 RSpec.describe Scrapers::HighPriority do
   describe '#perform' do
-    it 'add specs' do
-      county = create(:county, name: 'Tulsa')
-      test_case = create(:court_case, closed_on: 7.days.ago, county: county)
+    # let(:test_case) {create(:court_case, closed_on: 7.days.ago, county: county)}
+    before(:each) do
+      @county =  create(:county, name: 'Tulsa')
+      @test_case = create(:court_case, closed_on: 7.days.ago, county: @county)
+      @data = nil
```
sgelbart, line 11: I don't think you'd actually be able to reference these variables anywhere, that's why we use `let!` instead
But looks much better.

sgelbart (approved):

PR #157, opened 2023-03-30
Add DOC to Roster Table

AspiringDev51:


spec/lib/tasks/roster_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let!(:test_roster) { create(:roster) }
+  let!(:false_profile) { create(:doc_profile, roster: test_roster) }
+  context 'There are profiles with roster ids and some without' do
+    it 'creates roster and attaches it to profile without roster' do
+      Rake::Task['roster:empty_doc_profiles'].invoke
+      test_id = test_profile.id
+      test_profile = Doc::Profile.find_by(id: test_id)
+      expect(test_profile.roster).not_to  be nil
+      expect(false_profile.roster_id).to  be test_roster.id
+      expect(Doc::Profile.where(roster_id: nil).count).to be 0
```
sgelbart, line 17: looks good! Let's try adding some field tests too? We haven't been doing those but I think we should start. That means really comparing every field to what's in the test_profile

sgelbart (approved):
approved after you add a couple more tests for the content

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_02_17_162941) do
+ActiveRecord::Schema.define(version: 2023_03_28_153746) do
```
sgelbart, line 13: can you remove the schema.rb file from your PR since there are no migration changes?
In general this should only be included if you created a migration.

spec/lib/tasks/roster_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+require 'rake'
+
+Rails.application.load_tasks
+
+RSpec.describe 'roster' do
+  let!(:test_profile) { create(:doc_profile, roster: nil) }
+  let!(:test_roster) { create(:roster) }
+  let!(:false_profile) { create(:doc_profile, roster: test_roster) }
+  context 'There are profiles with roster ids and some without' do
```
sgelbart, line 10: These should be multiple specs
Please read https://www.betterspecs.org/

PR #170, opened 2023-05-11
Automate DOC import and add tests

sgelbart:
- Scrape files for DOC to AWS for import
- Validate schema hasn't changed by comparing readme
- Convert file names to be lowercase to avoid case issues
- Add simple tests


.gitignore
```diff
@@ -31,6 +31,9 @@ oscn.tar
 # Ignore Data Files
 /lib/data/*
 
+# Ignore large cassetts for testing
+/spec/cassettes/large
```
sgelbart, line 35: Added some DOC E2E tests (which can be recorded as cassettes)
Not actually committing the cassettes though since they're large, and skipping the tests by default.

spec/services/importers/doc/alias_spec.rb
```diff
@@ -1,9 +1,22 @@
 require 'rails_helper'
+require 'services/importers/doc/shared_specs'
 
 RSpec.describe Importers::Doc::Alias do
-  describe '#perform' do
-    it 'imports the data' do
-      skip
+  it_behaves_like('doc_importer') do
```
sgelbart, line 5: see shared_specs.rb

lib/tasks/doc.rake
```diff
@@ -8,69 +8,18 @@ require 'open-uri'
...
+    Scrapers::Doc::QuarterlyData.perform(dir)
+    ActiveRecord::Base.transaction do
+      Importers::Doc::OffenseCode.new(dir).perform
+      Importers::Doc::Profile.new(dir).perform
+      Importers::Doc::Sentence.new(dir).perform
+      Importers::Doc::Alias.new(dir).perform
+      Importers::Doc::Status.new(dir).perform
+      # Importers::Doc::LinkCases.new(county).perform # uncommented once we add support for multiple counties
+      # Importers::Doc::LinkOffenseCodes.new.perform # uncomment once this is confirmed to be working
+    end
```
sgelbart, line 23: Is this okay? Removed all the old commands to try to make a clear single entrypoint. If we need to run individual ones can always do it via console or one offs?

app/services/importers/doc/alias.rb
```diff
@@ -4,7 +4,7 @@ class Alias
       attr_accessor :file, :fields, :field_pattern, :bar, :doc_mapping, :dir
 
       def initialize(dir)
-        @file = Bucket.new.get_object("doc/#{dir}/Vendor_Alias_Extract_Text.dat")
+        @file = Bucket.new.get_object("doc/#{dir}/vendor_alias_extract_text.dat")
```
sgelbart, line 7: converting these to lowercase in quarterly_data.rb to deal with inconsistent casing (they're frequently suffixed with .DAT)

spec/services/importers/doc/alias_spec.rb
```diff
@@ -1,9 +1,22 @@
...
+    let(:class_to_import) { ::Doc::Alias }
+    let(:sample_file) do
+      ' 0000123482BELTON                        RICHARD                       LEWIS                         JR'
+    end
```
holdenmitchell, line 9: What would you think about moving this to either a Fixture or a Factory? It is mainly for aesthetics. But I could also see a scenario where a single character is deleted and it is hard to find the bug.

spec/services/importers/doc/offense_code_spec.rb
```diff
@@ -1,9 +1,21 @@
...
+require 'services/importers/doc/shared_specs'
 
 RSpec.describe Importers::Doc::OffenseCode do
-  describe '#perform' do
-    it 'imports the data' do
-      skip
+  it_behaves_like('doc_importer') do
+    let(:class_to_import) { ::Doc::OffenseCode }
+    let(:sample_file) do
+      '10-1144                               ACTS CAUSING JUVENILE DELINQUENCY       N'
```
holdenmitchell, line 8: Same as comment in the `alias_spec`

holdenmitchell (approved):
This looks good 👍 I appreciate all of the testing work you did and like the way you implemented it as a shared spec.

I think my main comments are around extracting the nasty sample file information from the tests for readability to either a factory or fixture file. Approving for the rush order 🍽️ 🏃 ✅

holdenmitchell (approved):
This looks good 👍 I appreciate all of the testing work you did and like the way you implemented it as a shared spec.

I think my main comments are around extracting the nasty sample file information from the tests for readability to either a factory or fixture file. Approving for the rush order 🍽️ 🏃 ✅

spec/services/importers/doc/alias_spec.rb
sgelbart: yeah good point..
I typically use factories but not sure how I feel about taking the factory and converting it in to the proper format (seems a bit clunky but might not be bad). When I use factories I'll reference the created value in my expected assertion. 
With fixtures you just hardcode the values even though they're coming from a separate file?

sgelbart: Resolving for now will try to come back to when I have more time!

PR #171, opened 2023-05-17
Add county_name arg to task

holdenmitchell:
# Description

Adds the arg `county_name` to the missing_cases rake task so that it can be used for any county. Also moves the task out of the one-off folder since it is proving to be useful


sgelbart (approved):
Thanks! I'm not sure you meant to keep the name of the worker as OneOffCaseWorker or not... (like it's scraping one cases so it kinda makes sense but that's confusing with the other "one-off", as in it will only run once)

PR #175, opened 2023-05-18
added enqueued field to party model, and its worker

AspiringDev51:


db/migrate/20230517212721_add_enqueue_to_parties.rb
```diff
@@ -0,0 +1,5 @@
+class AddEnqueueToParties < ActiveRecord::Migration[6.0]
+  def change
+    add_column :parties, :enqueued, :boolean
```
holdenmitchell, line 3: For the boolean fields, we generally want to set a default to prevent three states of a boolean column (`null`, `true` and `false`). In this case we would want it to be `false`. You can read more about [it here](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem)

```suggestion
    add_column :parties, :enqueued, :boolean, default: false
```

holdenmitchell (changes requested):
This looks good! I appreciate the testing that you have added to the specs. I've got one change that needs to be made with the database before we ship it 🚢

holdenmitchell (approved):
✅ LGTM

PR #202, opened 2023-08-30
 new branch for generated columns

AspiringDev51:


config/database.yml
```diff
@@ -74,7 +74,7 @@ test:
 production:
   <<: *default
   host: <%= ENV['DATABASE_HOST'] %>
-  database: postgres
+  database: postgres_
```
sgelbart, line 77: be sure to remove this before you're done

spec/models/court_case_spec.rb
```diff
@@ -68,6 +68,41 @@
     end
   end
 
+  describe 'generated clean_case_number column' do
```
sgelbart, line 71: you can reuse the same specs for all of these
https://rspec.info/features/3-12/rspec-core/example-groups/shared-examples/
the only difference should be the model you're creating and the name of the case number column. You can make thoes variables using lets.

PR #220, opened 2023-12-06
created extraction class

AspiringDev51:
Purpose: take date.month and year parameters to grab the correct url ex(https://www.okc.gov/departments/fire/daily-structure-fires/2023/july) and then use the date parameter to search the webpage and click on link to begin extracting  data from the pdf. returns json


holdenmitchell:
Part of #222

PR #223, opened 2023-12-18
Jw issue 1575

AspiringDev51:


app/services/importers/doc/profile.rb
```diff
@@ -62,9 +62,9 @@ def parse_status(data)
 
       def field_spacing(dir)
         if parse_date("#{dir}-01") > parse_date('2021-12-31')
-          [11, 30, 30, 30, 5, 8, 40, 8, 1, 40, 40, 2, 2, 4, 40, 10]
+          [10, 30, 30, 30, 4, 8, 50, 8, 1, 60, 60, 1, 2, 3, 60, 10]
         else
-          [11, 30, 30, 30, 5, 9, 40, 9, 1, 40, 40, 2, 2, 4, 40, 10]
+          [10, 30, 30, 30, 4, 9, 50, 8, 1, 60, 60, 1, 2, 3, 60, 10]
         end
```
sgelbart, line 68: so...this is something we really should be doing for all..
we should be keeping the spacing for the old files too in case we need to parse historic.
That means that we need to check the date and provide different formatting for each ones.
We also have to have the different versions in the tests.
I don't think we've been really doing this historically so we won't worry about it for now I guess, but that's someting to be aware of
Basically... We should assume that we do not support re-parsing of historic files!!!!

app/services/scrapers/doc/actual_readme.txt
```diff
@@ -0,0 +1,83 @@
+The data files are fixed width text files.
```
sgelbart, line 1: you need to replace the expected readme with this, shouldn't be it's own file

sgelbart (approved):

holdenmitchell (approved):
👍 LGTM

PR #249, opened 2024-07-05
Secretary of State

sgelbart:
```
Importers::OkSos::SplitFiles.perform('CORP_MSTR_240504.zip')
```


app/services/importers/ok_sos/split_files.rb
```diff
@@ -0,0 +1,121 @@
+module Importers
+  module OkSos
+    class SplitFiles < ApplicationService
```
sgelbart, line 3: this file should be split into two...one for splitting and one for importing.
I'm using tmp files which means keeping the @csvs variable in memory is important and splitting might give you issues
tbh though I think if you use a spawned dyno all files behave as temp files? Idk..

app/services/importers/ok_sos/split_files.rb
```diff
@@ -0,0 +1,121 @@
...
+          :officers
+        ].each do |file|
+          klass = Object.const_get("::Importers::OkSos::#{file.to_s.classify.pluralize}")
+          puts "importing #{klass}"
+          klass.perform(@csvs[file.to_s].path)
+        end
+      end
+
+      def split_files
+        columns = nil
```
sgelbart, line 68: this is extra complex because it's reading the file row by row but some rows are broken up because of new line characters so need to be "merged"
I'm sure there's a better way to handle but I couldn't find...

app/services/importers/ok_sos/split_files.rb
```diff
@@ -0,0 +1,121 @@
...
+          :suffixes,
+          :stock_types,
+          :entities,
+          :corp_filings,
+          :names,
+          :stock_data,
+          :stock_infos,
+          :agents,
+          :associated_entities,
+          :officers
```
sgelbart, line 59: if we order the "map" above we can avoid this

app/models/ok_sos/agent.rb
```diff
@@ -0,0 +1,5 @@
+class OkSos::Agent < ApplicationRecord
+  belongs_to :suffix, class_name: 'OkSos::Suffix', optional: true
+  belongs_to :entity_address, class_name: 'OkSos::EntityAddress', optional: true
+  belongs_to :entity, class_name: 'OkSos::Entity', optional: true
+end
```
sgelbart, line 5: still could do reverse of the belongs_to

spec/services/importers/tulsa_blotter/daily_import_spec.rb
```diff
@@ -11,10 +11,10 @@
           Importers::TulsaBlotter::DailyImport.perform # NOTE: vcr has been shortened
         end
 
-        expect(::TulsaBlotter::PageHtml.count).to be > 0
-        expect(::TulsaBlotter::Arrest.count).to be > 0
-        expect(::TulsaBlotter::ArrestDetailsHtml.count).to be > 0
-        expect(::TulsaBlotter::Offense.count).to be > 0
+        expect(TulsaBlotter::PageHtml.count).to be > 0
```
sgelbart, line 14: idk why rubo is doing this...

spec/services/importers/okc_blotter/pdf_spec.rb
```diff
@@ -10,9 +10,9 @@
         ENV['BUCKETEER_BUCKET_NAME'] = 'bucket-name' # aws s3 changes behavior if bucket is uppercase or has underscores
         VCR.use_cassette 'scrape_pdf' do
           Importers::OkcBlotter::Pdf.perform('2022-10-20')
-          expect(::OkcBlotter::Pdf.count).to be > 0
-          expect(::OkcBlotter::Booking.count).to be > 0
-          expect(::OkcBlotter::Offense.count).to be > 0
+          expect(OkcBlotter::Pdf.count).to be > 0
```
sgelbart, line 13: need to fix rubo... or something...

PR #253, opened 2024-08-21
SOS Performance Fixes

sgelbart:
# Summary

Cut the local runtime from 5+ hours to closer to two.
The changes may be more noticeable on the server as it has to go over the wire for the database.

# Changes

**1) Insert ZipCodes for addresses in bulk**
This was being done row by row and due to the large amount of zip codes was still taking some time

**2) Cache larger lookups in memory.** 
The variable containing the large lookup (1390000 records) seem to be pretty well optimized in memory but take a few minutes to load. 

Testing info:

  ```
irb(main):002:0> Time.now
=> 2024-08-21 18:53:31.129373628 +0000
irb(main):003:0> test = ::OkSos::Entity.pluck(:filing_number, :id).to_h
D, [2024-08-21T18:53:40.920200 #2] DEBUG -- :   OkSos::Entity Pluck (6723.8ms)  SELECT "ok_sos_entities"."filing_number", "ok_sos_entities"."id" FROM "ok_sos_entities"
=>                                                                                                                                     
{1912071632=>630019,                                                                                                                   
...                                                                                                                                    
irb(main):004:0> Time.now
=> 2024-08-21 18:56:45.623127182 +0000
  ```
 
  ```
  irb(main):007:0> ObjectSpace.memsize_of(test)
=> 67108960
irb(main):008:0> puts 'RAM USAGE: ' + `pmap #{Process.pid} | tail -1`[10,40].strip
RAM USAGE: 1345428K
  ```

Tested on performance-2xl dyno

### DIFF
--- a/app/services/importers/ok_sos/base_importer.rb
+++ b/app/services/importers/ok_sos/base_importer.rb
@@ -9,7 +9,7 @@ module Importers
 
       def initialize(file_path)
         @file_path = file_path
-        @@model_cache = ActiveSupport::HashWithIndifferentAccess.new({})
+        @@id_cache = ActiveSupport::HashWithIndifferentAccess.new({})
         super()
       end
 
@@ -85,15 +85,15 @@ module Importers
         end
       end
 
-      def model_cache(klass, key)
+      def id_cache(klass, key)
         klass_key = klass.to_s
-        return @@model_cache[klass_key] if @@model_cache[klass_key]
+        return @@id_cache[klass_key] if @@id_cache[klass_key]
 
-        @@model_cache[klass_key] = klass.pluck(key, :id).to_h { |x| [x[0].to_s, x[1]] }
-        @@model_cache[klass_key]
+        @@id_cache[klass_key] = klass.pluck(key, :id).to_h { |x| [x[0].to_s, x[1]] }
+        @@id_cache[klass_key]
       end
 
-      def get_cached(klass, key, value, create: false)
+      def lookup_cached_id(klass, key, value, create: false)
         return nil unless value.present? && value != '0'
 
         return model_cache(klass, key)[value.to_s] if model_cache(klass, key)[value.to_s]
@@ -101,7 +101,7 @@ module Importers
         raise ActiveRecord::RecordNotFound unless create
 
         new_model = klass.create!(key => value)
-        @@model_cache[klass.to_s][value.to_s] = new_model.id
+        @@id_cache[klass.to_s][value.to_s] = new_model.id
         new_model.id
       end
 
--- a/app/services/importers/ok_sos/base_importer.rb
+++ b/app/services/importers/ok_sos/base_importer.rb
@@ -96,7 +96,7 @@ module Importers
       def lookup_cached_id(klass, key, value, create: false)
         return nil unless value.present? && value != '0'
 
-        return model_cache(klass, key)[value.to_s] if model_cache(klass, key)[value.to_s]
+        return id_cache(klass, key)[value.to_s] if id_cache(klass, key)[value.to_s]
 
         raise ActiveRecord::RecordNotFound unless create
 
--- a/app/services/importers/ok_sos/base_importer.rb
+++ b/app/services/importers/ok_sos/base_importer.rb
@@ -32,7 +32,8 @@ module Importers
           rescue StandardError => e
             error_rows << { row: row, e: e }
           end
-          if ((rows + error_rows).count % BATCH_SIZE).zero? || rows.count == row_count
+          total_rows = (rows + error_rows).count
+          if (total_rows % BATCH_SIZE).zero? || total_rows == row_count
             rows = check_and_fix_duplicates(rows)
             bar&.increment!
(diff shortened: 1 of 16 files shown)


app/services/importers/ok_sos/base_importer.rb
```diff
@@ -9,7 +9,7 @@ class BaseImporter < ApplicationService
 
       def initialize(file_path)
         @file_path = file_path
-        @model_cache = ActiveSupport::HashWithIndifferentAccess.new({})
+        @@model_cache = ActiveSupport::HashWithIndifferentAccess.new({})
```
sgelbart, line 12: We have to be a bit careful about this because once the cache is generated no changes will be reflect even between the different importer files

app/services/importers/ok_sos/base_importer.rb
```diff
@@ -32,7 +32,8 @@ def do_import
           rescue StandardError => e
             error_rows << { row: row, e: e }
           end
-          if ((rows + error_rows).count % BATCH_SIZE).zero? || rows.count == row_count
+          total_rows = (rows + error_rows).count
+          if (total_rows % BATCH_SIZE).zero? || total_rows == row_count
```
sgelbart, line 36: little bug fix for if there's errors

app/services/importers/ok_sos/base_importer.rb
```diff
@@ -85,11 +86,11 @@ def parse_date(date)
...
       def model_cache(klass, key)
-        cache_key = klass.to_s
-        return @model_cache[cache_key] if @model_cache[cache_key]
+        klass_key = klass.to_s
+        return @@model_cache[klass_key] if @@model_cache[klass_key]
 
-        @model_cache[cache_key] = klass.all.to_h { |x| [x[key].to_s, x] }
-        @model_cache[cache_key]
+        @@model_cache[klass_key] = klass.pluck(key, :id).to_h { |x| [x[0].to_s, x[1]] }
+        @@model_cache[klass_key]
```
sgelbart, line 93: optimize in memory since we're always just referencing id

app/services/importers/ok_sos/base_importer.rb
```diff
@@ -84,24 +90,24 @@ def parse_date(date)
...
 
-      def get_cached(klass, key, value, create: false)
+      def lookup_cached_id(klass, key, value, create: false)
         return nil unless value.present? && value != '0'
 
-        return model_cache(klass, key)[value.to_s] if model_cache(klass, key)[value.to_s]
+        return id_cache(klass, key)[value.to_s] if id_cache(klass, key)[value.to_s]
 
-        raise ActiveRecord::RecordNotFound unless create
+        return nil unless create
```
sgelbart, line 106: Return nil instead of raising an error if the record isn't found. Since we're storing external ids to the table it's a bit better here not to raise an error so we can still create records and then see what ids are missing if that makes sense (this will only work if the association is nullable or else will fail anyhow)

app/services/importers/ok_sos/base_importer.rb
```diff
@@ -32,10 +32,16 @@ def do_import
...
           end
-          if ((rows + error_rows).count % BATCH_SIZE).zero? || rows.count == row_count
+          total_rows = (rows + error_rows).count
+          if (total_rows % BATCH_SIZE).zero? || total_rows == row_count
+            print_errors(error_rows) if error_rows.present?
             rows = check_and_fix_duplicates(rows)
             bar&.increment!
-            import_class.upsert_all(rows, unique_by: unique_by)
+            if rows.blank?
+              puts "empty list of rows "
```
sgelbart, line 41: There's an issue where the last 20k+ rows of corp_filings.csv have filling numbers for entities that aren't in entities.csv, so a whole batch could be empty. Did another fix for this below too.

sgelbart: Also, I wrote with `puts` assuming we'd be monitoring the run, but depending upon the final run might be better to switch to Raygun.

app/services/importers/ok_sos/base_importer.rb
```diff
@@ -9,7 +9,7 @@ class BaseImporter < ApplicationService
 
       def initialize(file_path)
         @file_path = file_path
-        @model_cache = ActiveSupport::HashWithIndifferentAccess.new({})
+        @@id_cache = ActiveSupport::HashWithIndifferentAccess.new({})
```
sgelbart, line 12: Have to be extra careful with the cache now since it persists across imports.
Did this since the initial load for Entities etc takes a while.

app/services/importers/ok_sos/split_files.rb
```diff
@@ -38,6 +38,8 @@ def perform
       end
 
       def import_files
+        ::Importers::OkSos::ZipCodes.perform(@csvs["entity_addresses"].path)
```
sgelbart, line 41: Loading zip codes from entity addresses saves some time on the first run. Thousands of zips are loaded from all over the country so there's quite a few.

holdenmitchell (approved):
👍

PR #254, opened 2024-09-04
OK Assessor Import

sgelbart:
Imports Oklahoma Assessor data.

To run use: `bundle exec rake 'assessor:import[2024-08-15]'`

Files downloaded as CSVs from "Layers" on 
https://ok-county-gis-hub-ok-co.hub.arcgis.com/maps/6cffa5c4514b49a28db2932312abb72b/about

Saved using using naming convention View_OKPublicRecordValueDetail.csv

Sample script for uploading files from S3 from local:

```ruby
directory = "~/Downloads/assessor-2024-08-15"
Dir.children("~/Downloads/assessor-2024-08-15").each do |file|
  puts "uploading #{file}"
  content = File.new("#{directory}/#{file}")
  Bucket.new.put_object("ok_assessor/2024-08-15/#{file}", content)
end
```

## Tips for review:
* Review base_importer.rb
* Review the importer specs to see how the final data comes in
* Review the unique_by keys (I did analysis on the sample dataset to determine these but we may want to update again based on the final one)

## Sample Query
```pgsql
-- see average sale price for each city (note: zipcode and a lot of address data is largely missing from the accounts file
select city, date_trunc('year', sale_date), avg(sale_price), count(*)
from (select sale_date, ok_assessor_accounts.city, sale_price
      from ok_assessor_accounts
               join ok_assessor_sales
                    on ok_assessor_accounts.id = ok_assessor_sales.account_id
      order by sale_date)
group by city, date_trunc('year', sale_date)
order by city, date_trunc('year', sale_date)
```


app/services/importers/ok_assessor/sales.rb
```diff
@@ -0,0 +1,47 @@
...
+        }
+      end
+
+      def validate_attributes!
+        raise AttributeError, "No account found for number: #{row['ACCOUNT_NUM']}" if attributes[:account_id].nil?
+      end
+
+      def unique_by
+        [:account_id, :reception_number]
+      end
```
sgelbart, line 32: In general there are a lot of unique_by constraints that still result with duplicates in batches. I confirmed that most of them are truly duplicates. This was probably the only one I considered tweaking.

My guess is that this key is largely correct but if a sale is attempted between two parties and fails i think a new one (between two different parties) might come through with the same reception number. I was seeing this for about 1/100 sales and currently they're being ignored as duplicates.

PR #256, opened 2024-09-04
Fix SOS

sgelbart:
Fix bug that prevented last batch of records from getting inserted if multiple batches present

### DIFF
--- a/app/services/importers/ok_sos/base_importer.rb
+++ b/app/services/importers/ok_sos/base_importer.rb
@@ -25,15 +25,16 @@ module Importers
         if row_count > BATCH_SIZE
           bar = ProgressBar.new(row_count / BATCH_SIZE)
         end
+        options = {col_sep: ',', quote_char: '"', headers: true, liberal_parsing: true}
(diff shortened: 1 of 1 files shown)


app/services/importers/ok_sos/base_importer.rb
```diff
@@ -25,15 +25,16 @@ def do_import
...
-          .foreach(file_path, col_sep: ',', quote_char: '"', headers: true, liberal_parsing: true) do |row|
+          .foreach(file_path, **options).with_index do |row, i|
           begin
             rows << attributes(row)
           rescue StandardError => e
             error_rows << { row: row, e: e }
           end
           total_rows = (rows + error_rows).count
-          if (total_rows % BATCH_SIZE).zero? || total_rows == row_count
+          if (total_rows % BATCH_SIZE).zero? || i+1 == row_count
```
sgelbart, line 37: I was reusing total rows here when that actually gets reset each batch insert
(so the last batch in this case wouldn't get inserted)

PR #260, opened 2024-09-19
Validate for breaking HTML changes

sgelbart:
Check all cases in the stored examples directory and make sure they parse to the expected json


Gemfile
```diff
@@ -19,7 +19,7 @@ gem 'mailgun-ruby', '~>1.2.14'
 gem 'monetize'
 gem 'ok2explore', git: 'https://github.com/AyOK-Code/ok2explore', ref: 'fdf184d'
 gem 'ocso_scraper', git: 'https://github.com/AyOK-Code/ocso_scraper', ref: '4023411'
-gem 'oscn_scraper', git: 'https://github.com/AyOK-Code/oscn_scraper', ref: 'f2904ac'
+gem 'oscn_scraper', git: 'https://github.com/AyOK-Code/oscn_scraper', ref: '31085ee'
```
sgelbart, line 22: parsing parties for SC cases was working from the site but not from a fixture (with matching HTML)
This seemed brittle enough to warrant a fix so it doesn't cause gotchas for anyone else.

sgelbart: I can't remember if that sha will persist once the scraper PR is merged in...might need to update

PR #263, opened 2024-10-31
Lexus Nexus

sgelbart:
- Add import from the NACOK website. This includes historic PDFs of OKCPD crimes.
- Add import from the community crime map website. This includes historic crimes for many different agencies but may have data gaps.

To run the historic scrapes use:
```
bundle exec rake 'lexus_nexus:import_from_crime_map'
```
(takes under 10 minutes running locally)
or 
```
bundle exec rake 'lexus_nexus:import_from_nacok'
```
(takes just under 2 hours running locally)

If you just want to test one PDF from nacok you can just pass the URL directly into the PDF scraper from console.


app/services/importers/crime_map.rb
```diff
@@ -0,0 +1,374 @@
...
+      {
+        'buffer' => {
+          'enabled' => false,
+          'restrictArea' => false,
+          'value' => []
+        },
+        'date' => {
+          'start' => '09/23/2024',
+          'end' => '10/23/2024'
+        },
```
sgelbart, line 67: oh snap I just noticed that

app/services/importers/crime_map.rb
```diff
@@ -0,0 +1,374 @@
...
+        'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36'
+      }
+      res = HTTParty.get(url, headers: headers)
+      @token = JSON.parse(res.body)['data']['jwt']
+      @token
+    end
+    # rubocop:enable Metrics/LineLength
+
+    # rubocop:disable Metrics/MethodLength
+    def selection
```
sgelbart, line 146: If I remove this whole thing the request fails..might be able to at least shorten

app/services/importers/crime_map.rb
```diff
@@ -0,0 +1,379 @@
...
+      @zoom = 9
+    end
+
+    def perform
+      date_range = (Date.parse(ENV.fetch('LEXUS_NEXUS_START', '2020-01-01'))..Time.zone.now)
+      date_range.step(3).each do |date|
+        @start_date = date
+        @end_date = date + 1.week
+        save_crimes
+      end
```
sgelbart, line 28: it turns out we can kinda get in historic using the crime map
The results I get are very inconsistent. For September we got in over a thousand distinct using the pdf but only got in a fraction of them using the crime map, but for some moths we get many many more.

sgelbart: I checked Moore, Norman, Midwest City, and Edmond after doing an import back to 2020 and they all look somewhat consistent compared to OKCPD, who jumps between 1000 to 100 to 30. So hopefully between the two imports we will have a decent dataset, but we could do more analysis.

sgelbart: There are also a lot of null `incident_at`s I could look into more
```pgsql
select agency, date_trunc('month', incident_at), count(*) from lexus_nexus2_crimes
group by agency, date_trunc('month', incident_at)
order by agency, date_trunc('month', incident_at) desc
```

spec/models/lexus_nexus/crime_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe LexusNexus::Crime, type: :model do
+  describe 'associations' do
+    it { should validate_presence_of(:incident_number) }
```
sgelbart, line 5: I don't have almost any tests...

app/services/importers/nacok_crime/pdf.rb
```diff
@@ -0,0 +1,115 @@
...
+          'Total incidents',
+          'Reporting Dates:'
+        ]
+        exclude_rows_with_text.any? { |text| row.include?(text) }
+      end
+
+      def columns(header_row)
+        cols = {}
+        # Address also includes Date and Time which are extracted using regex
+        column_names = ['Address', 'Offense', 'Description', 'Division', 'Case Number']
```
sgelbart, line 91: so this is a bit wonky because technically there are columns for date and time but I just through them in here with Address because I have better results to extracting them from that using regex

app/models/lexus_nexus/crime.rb
```diff
@@ -0,0 +1,10 @@
+class LexusNexus::Crime < ApplicationRecord
```
holdenmitchell, line 1: Going to adjust this namespace to be independent of Lexus Nexus since we may or may not get to use that data.

holdenmitchell (approved):
Made a few small updates:

1. Turned off the ability to import Lexus Nexis because of some concerns around their terms & conditions. We are trying to get in touch with them.
2. Changed the namespace to be under Community instead of Lexus Nexis.
3. Added `source_data` field to log where the data came from

I'm going to go ahead and deploy and pull in the PDF data. Need to create a follow up PR on the monthly automation.

PR #269, opened 2025-01-29
Prevent record cap from being hit on Ok2Explore

sgelbart:
## scrape:death_records  Updates
For scrape:death_records in deaths.rake
- break up into multiple importers
- fan out jobs that hit limit into smaller jobs
- utilize progress bar with progress bar logging
- skip records with no county

## scrape:jobs Updates
For scrape:jobs in deaths.rake
- pass in a year to create scrape jobs, e.g., `bundle exec rake 'scrape:jobs[2019]`

### DIFF
--- a/app/models/ok2_explore/death.rb
+++ b/app/models/ok2_explore/death.rb
@@ -1,5 +1,5 @@
 class Ok2Explore::Death < ApplicationRecord
-  belongs_to :county
+  belongs_to :county, optional: true
 
   validates :first_name, :last_name, :death_on, presence: true
   enum sex: [:male, :female]
--- a/app/services/importers/ok2_explore/scrape_job.rb
+++ b/app/services/importers/ok2_explore/scrape_job.rb
@@ -20,7 +20,6 @@ module Importers
           return nil
         end
 
-        log "Found #{death_records.count} scrape_jobs."
         import_death_records(death_records)
 
         scrape_job.update(is_success: true)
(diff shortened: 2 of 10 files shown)


app/models/ok2_explore/death.rb
```diff
@@ -1,5 +1,5 @@
 class Ok2Explore::Death < ApplicationRecord
-  belongs_to :county
+  belongs_to :county, optional: true
```
sgelbart, line 2: When testing I made it about 1/5 of the way through the year and noticed a handful of records with no county. Went ahead and made nullable but that's in its own commit if we want to revert.

app/services/importers/ok2_explore/scrape_job.rb
```diff
@@ -0,0 +1,69 @@
...
+        super()
+      end
+
+      def perform
+        log "Scraping: #{scrape_job.id}"
+        begin
+          death_records = ::Ok2explore::Scraper.new(**scraper_args).perform
+        rescue ::Selenium::WebDriver::Error::TimeoutError
+          log 'Timeout occurred, skipping for now.'
+          return nil
```
sgelbart, line 18: I hit the timeout once but when you're in the middle of a big run it's frustrating. If we want to switch this back to being in the worker we can do that too which might handle better.

holdenmitchell (approved):
🚢 it!

PR #274, opened 2025-05-09
KP Counties

sgelbart:


app/services/scrapers/case.rb
```diff
@@ -10,8 +10,16 @@ def perform
...
+        if OscnScraper::Requestor::DAILING_FILINGS_COUNTIES.include? county
+          Scrapers::NewCases.perform(county, days_ago: days_ago)
+          Scrapers::HighPriority.perform(county, days_ago: days_ago, days_forward: days_forward)
+        else
+          date_range = (DateTime.now.to_date - days_ago...DateTime.now.to_date)
+          date_range.each do |date|
+            Scrapers::DailyFilingsAltCounties.perform(county, date, enqueue: false)
+          end
+          Scrapers::HighPriority.perform(county, include_event_updates: false)
+        end
```
holdenmitchell, line 22: It might easier to understand if you use the counties table scope `counties.ocis` and `counties.kp` rather than a constant outside of the project.

```ruby
enum rms: {
    ocis: 'OCIS',
    kp: 'KP'
}
```

app/services/importers/party_html.rb
```diff
@@ -13,8 +13,10 @@ def self.perform(oscn_id)
     end
 
     def perform
-      county_name = @party.court_cases.first.county.name
-      data = OscnScraper::Requestor::Party.fetch_party(county_name, oscn_id)
+      sample_courtcase = @party.court_cases.first
+      county_name = sample_courtcase.county.name
+      case_number = sample_courtcase.case_number
+      data = OscnScraper::Requestor::Party.fetch_party(county_name, oscn_id, case_number)
```
holdenmitchell, line 19: This assumes that Parties never cross counties. I think it is a safe assumption, but I don't remember ever checking after we expanded beyond OK county

app/models/county.rb
```diff
@@ -10,6 +10,11 @@ class County < ApplicationRecord
 
   validates :name, :fips_code, presence: true
 
+  enum rms: {
+    ocis: 'OCIS',
+    kp: 'KP'
+  }
+
```
holdenmitchell, line 17: Is this a non-numeric enum? I'm going to start using this immediately. So much better than numeric mapping

db/other_sql/clean_case_number_v01.sql
```diff
@@ -0,0 +1,39 @@
+-- replace crf_number with any case number field
```
holdenmitchell, line 1: I'm not sure where this is used, but just so you know, clean case number already exists on both `doc_sentences` and `ocso_warrants` as generated columns. See these files:

- `20230621153902_add_clean_case_number_to_ocso_warrants`
- `20240710130156_add_clean_case_number_to_doc_sentences`

If I remember correctly, both SQL statements could be improved a bit

holdenmitchell (approved):
Looks great! Comments are for you to take or leave, but definitely check out the one with about the clean case number columns

holdenmitchell:
@sgelbart I will work on getting all the views adjusted to only include the OCIS cases.

app/services/importers/event.rb
```diff
@@ -47,6 +47,7 @@ def find_event(event_data)
         .find_or_initialize_by({
                                  court_case_id: court_case.id,
                                  event_at: event_data[:date],
```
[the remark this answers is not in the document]
sgelbart: allow for mutliple events on the same day with different names

app/services/importers/party_html.rb
sgelbart: I don't think it would actually matter if there were because I'd assume the party data for all counties would be the same.

PR #322, opened 2025-09-25
Hm/sex offenders scraper

holdenmitchell:
# Description


app/models/ok_sex_offender/address.rb
```diff
@@ -0,0 +1,8 @@
+class OkSexOffender::Address < ApplicationRecord
+  belongs_to :offender
+
+  validates :address_type, presence: true
+
+  scope :home, -> { where(address_type: 'HOME') }
```
sgelbart, line 6: are all these scopes etc in use anywhere?

